PR #6, opened 2024-09-18
Correct typos in Specs markdown

LL782:
I noticed a few typos while reading the specs


lcastillo-ledger:
Thanks! typos have been fixed in another commit

PR #29, opened 2024-10-07
ci: use `erc7730` from PyPI instead of internal repository

jnicoulaud-ledger:
see LedgerHQ/python-erc7730#57


amanone:
Hello 👋 
Feedback with @AEnguerrand : 
on `master.yaml` we suggest to keep the pull from JFrog
`run: jf pip install erc7730`

but change
`run: jf pipc --global --repo-resolve=virtual-pypi-prod-green`
to 
`run: jf pipc --global --repo-resolve=ossproxy-pypi-proxy-pypiorg`

and on the `workflow pull_request.yaml`, indeed, we should pull directly from PyPI 🙏

jnicoulaud-ledger:
@amanone @AEnguerrand it does not work from forks, that's why we are switching to PyPI, see https://github.com/LedgerHQ/clear-signing-erc7730-registry/actions/runs/11184089433/job/31094107439?pr=18

AEnguerrand:
> @amanone @AEnguerrand it does not work from forks, that's why we are switching to PyPI, see https://github.com/LedgerHQ/clear-signing-erc7730-registry/actions/runs/11184089433/job/31094107439?pr=18

Yes, on the workflow running for PE, definitely don't use JFrog. But for the workflow master, running only on the master branch, you can use JFrog (and modify it); I think it's not going to break in this case. It will be beneficial to have the JFrog Xray security check.

jnicoulaud-ledger:
@AEnguerrand If you fork the repo, you may want to have the master workflow working on your fork. I could see this happening easily with third parties managing multiple descriptors before submitting a big PR to us. Also, pull requests and master won't source the library from the same location, so might lead to weird behaviour if release fails to deploy to PyPI or JFrog. So I will put back if you want, but otherwise I would prefer to keep it simple

amanone (approved):
_removed_
@AEnguerrand please provide status. thanks :)

amanone (approved):
Thanks for the update @jnicoulaud-ledger !

jnicoulaud-ledger:
reverted master workflow because it can't work on a fork anyway, due to using ledger runners

lcastillo-ledger (approved):

PR #37, opened 2024-10-09
chore: add prototools file

jnicoulaud-ledger:


developer-preview/.prototools
```diff
@@ -0,0 +1 @@
+node = "18.17.0"
```
LL782: Let's use the LTS (v20.18.0) see: https://nodejs.org/en

```suggestion
node = "20.18.0"
```

LL782 (approved):
Great suggestion. I actually meant to add an `.nvmrc` with version v20.18

I'm not familiar with prototools but the principle looks the same.

LL782 (approved):

PR #42, opened 2024-10-09
demo laurent

lcastillo-ledger:
Add demo file


jnicoulaud-ledger (approved):

jnicoulaud-ledger:
@lcastillo-ledger it didn't work because your file is not named `eip712-*.json`, we have naming convention in place because it's hard to differentiate actual descriptors from partial include files

PR #61, opened 2024-10-18
fix: fix `@.amount`/`@.value` confusion

jnicoulaud-ledger:
fix "native currency value of the transaction" sometimes referred as `@.amount`, sometimes `@.value` => use `@.value` everywhere


fsamier (approved):

jnicoulaud-ledger:
@lcastillo-ledger I am force merging because it is blocking our tests, feel free to revert

PR #63, opened 2024-10-21
spec: allow more data types in `$.metadata.constants` section

jnicoulaud-ledger:
note: I did not enable arrays/objects (yet), as implementation of resolution is a bit more complex


fsamier (approved):

jnicoulaud-ledger:
@lcastillo-ledger fyi, applying this spec change to not be blocked, we can revert if you disagree

PR #86, opened 2024-11-16
Add ERC4626 & Kiln DeFi registry

kilnmg:
Added ERC4626 in the erc folder: 
- https://eips.ethereum.org/EIPS/eip-4626
Todo: 
Need to find a better way to make it generic over multiple deployments as of today it's looks like it's always getting : `.deployments.[0]` while with ERC4626 DeFi protocols will have a lot of vaults to add in the registry. 
```
    "params": {
        "tokenPath": "$.context.contract.deployments.[0].address"
    }
```

Added a new registry for Kiln DeFi contracts: 
- https://docs.kiln.fi/v1/kiln-products/defi/security/source-code 
- Only including an Ethereum Mainnet Vault and a Sepolia testnet vault.

Preview: 
ERC4626-Deposit

Opening a PR for a more advanced review on : http://app.devicesdk.ledger-test.com/

### DIFF
--- a/registry/starkgate/calldata-StarkGate-STRK.json
+++ b/registry/starkgate/calldata-StarkGate-STRK.json
@@ -17,9 +17,9 @@
         "$id": "deposit",
         "intent": "Bridge",
         "fields": [
-          { "path": "token", "label": "Token", "format": "addressName", "params": { "types": ["token"] } },
-          { "path": "amount", "label": "Amount", "format": "amount" },
-          { "path": "l2Recipient", "label": "Recipient", "format": "raw" }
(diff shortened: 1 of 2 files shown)


jnicoulaud-ledger (commented):
I understand what you are trying to do, but unfortunately it's not possible in the current form:
 - constants are restricted to `$.metadata.constants`
 - extra fields in `deployments` are not allowed
 - referencing a value "for current deployment" is not possible

I think you can achieve what you want this way:
 - in the ERC file, define your display section, using `$.metadata.constants.asset`, and other constant stuff such as ABI
 - then create one calldata file per deployment, it will only contain deployments section + metadata.constants.asset

PR #99, opened 2024-11-27
feat: BACK-8066  One deployment per chain + exhaustive list of selectors

ClaireGuerreGiordano:
This pr aims to list only deployment of interest within paraswap descriptor. It also list all known selectors

[JIRA ticket](https://ledgerhq.atlassian.net/browse/BACK-8066)


registry/paraswap/calldata-AugustusSwapper-v5.json
```diff
@@ -4,12 +4,7 @@
     "contract": {
       "abi": "https://github.com/LedgerHQ/ledger-asset-dapps/blob/main/ethereum/paraswap/abis/0xdef171fe48cf0115b1d80b88dc8eab59176fee57.abi.json",
       "deployments": [
-        { "chainId": 1, "address": "0x1bd435f3c054b6e901b7b108a0ab7617c808677b" },
-        { "chainId": 1, "address": "0xdef171fe48cf0115b1d80b88dc8eab59176fee57" },
-        { "chainId": 56, "address": "0x55a0e3b6579972055faa983482aceb4b251dcf15" },
-        { "chainId": 56, "address": "0xdef171fe48cf0115b1d80b88dc8eab59176fee57" },
-        { "chainId": 137, "address": "0x90249ed4d69d70e709ffcd8bee2c5a566f65dade" },
-        { "chainId": 137, "address": "0xdef171fe48cf0115b1d80b88dc8eab59176fee57" }
+        { "chainId": 1, "address": "0xdef171fe48cf0115b1d80b88dc8eab59176fee57" }
```
jnicoulaud-ledger: Why remove those on chain id 56 and 137, are they pointing to different implementation ?

jnicoulaud-ledger (dismissed):

registry/paraswap/calldata-AugustusSwapper-v5.json
ClaireGuerreGiordano: I thought you wanted only the current descriptor related deployment in the file. Plus, which one should I keep? As you can see, for chain id 56 and 137, there are also 2 address on 1 chain.

jnicoulaud-ledger: If contract has same address on a different chain, then it is the same contract because address is hash of contract contents. This is what we call a deployment, same contract deployed on a specific chain. Then the same descriptor can be applied since it the same ABIs.

**unless edge case**: contract is only a proxy redirecting to an implementation contract. In this case it is possible that contract has same address, but a different implementation on different chains. This is what you need to verify to know if descriptor can be applied to all chains.

ClaireGuerreGiordano: As you stated, all contract with address 0xdef171fe48cf0115b1d80b88dc8eab59176fee57 belongs to paraswap v5 contract, and there is only one deployment for paraswap v4 on chain id 1 (0x1bd435f3c054b6e901b7b108a0ab7617c808677b). Fixed as you suggested.

PR #104, opened 2024-12-01
✨ developer-preview: decode transaction calldata

### LINKED ISSUE #78: add tx simulation
ideally I would like to pass some tx ids and see what Flex would have displayed

cruzdanilo:
# changes
- receive hex calldata as input for preview values
- fetch url-based abis, handle special cases: non-raw github, etherscan api (requires key through env var `NEXT_PUBLIC_ETHERSCAN_API_KEY`, fails gracefully without it)
- decode calldata based on abi, fallback to format signature when human-readable
- format timestamps, units and token amounts
- resolve ens names (reverse lookup)
- add viem dependency, and a polyfill for `TextEncoder` (only for jest tests)
- closes #78

Screenshot 2024-12-01 at 13 38 14


shigekobeasley (approved):

LL782:
@cruzdanilo thank you very much for your PR and apologies that I missed the notification before now. After a first read of the description and a first scan of the code changes this looks like excellent work. @pdeville-ledger and/or myself will try to give it a full review this week.

Firstly, I'd like to share some context around the preview tool. This locally run web-app is very likely to be replaced and removed from this repository. We just started work on a hosted web-app that will generate the Clear-Signing JSON file, starting with a contract address or ABI and providing UI for adding the Clear-signing meta-data. The preview will be live in the browser and the end result will be a JSON file to add to this repository (the registry). So while some of the functionality will be the same as this preview tool we want to be transparent that the preview tool will probably be removed from this repository and much of the code will be replaced.

fsamier:
Closed against https://github.com/LedgerHQ/clear-signing-erc7730-registry/pull/144
Thanks for your contribution! However developer tools in this repository were a temporary solution, and thus have been removed. More details on tooling will be communicated once available.

PR #108, opened 2024-12-09
Dynamic preview

2manslkh:
Add dynamic preview


fsamier:
Closed against https://github.com/LedgerHQ/clear-signing-erc7730-registry/pull/144
Thanks for your contribution! However developer tools in this repository were a temporary solution, and thus have been removed. More details on tooling will be communicated once available.

PR #873, opened 2025-03-19
add common celo core contract operations

aaronmgdr:
Adds clear signing for calls that can be made against celo core contracts by regular accounts.

### DIFF
--- a/registry/celo/calldata-Accounts-v1.1.4.2.json
+++ b/registry/celo/calldata-Accounts-v1.1.4.2.json
@@ -1,0 +1,2376 @@
+{
+    "$schema": "../../specs/erc7730-v1.schema.json",
+    "context": {
+      "$id": "Celo Accounts",
+      "contract": {
+        "deployments": [
+          {
+            "chainId": 42220,
+            "address": "0x7d21685C17607338b313a7174bAb6620baD0aaB7"
+          },
+          {
+            "chainId": 44787,
+            "address": "0xed7f51A34B4e71fbE69B3091FcF879cD14bD73A9"
+          }
+        ],
+        "abi": [
+          {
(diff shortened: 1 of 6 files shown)


registry/celo/calldata-Accounts-v1.1.4.2.json
```diff
@@ -0,0 +1,2376 @@
...
+              "path": "#.beneficiary",
+              "value": null
+            },
+            {
+              "$id": null,
+              "label": "Fraction",
+              "format": "unit",
+              "params": {
+                "base": "%",
+                "decimals": "",
```
jnicoulaud-ledger, line 2326: This is where the linter is not happy, `decimals` is mandatory, eg:

```suggestion
                "decimals": 0,
```

registry/celo/calldata-Election-v1.1.4.0.json
```diff
@@ -0,0 +1,2723 @@
...
+              "value": null
+            },
+            {
+              "$id": null,
+              "label": "Value",
+              "format": "unit",
+              "params": {
+                "base": "",
+                "decimals": "18",
+                "prefix": "CELO"
```
jnicoulaud-ledger, line 2518: `prefix` is a boolean:

> A boolean indicating whether an SI prefix should be appended, defaults to False

I think you are looking for `base`: https://github.com/LedgerHQ/clear-signing-erc7730-registry/blob/master/specs/erc-7730.md?plain=1#L1012

jnicoulaud-ledger (commented):
Fyi, you can install the linter tool locally to iterate more easily: https://ledgerhq.github.io/python-erc7730/pages/usage_cli.html

PR #1116, opened 2025-04-01
Add SwissBorg

antoineborg:
SwissBorg contracts:
- `ChsbToBorgMigrator`, Migrator contract to get BORG from CHSB tokens, see https://docs.swissborg.com/borg-migration/contract-addresses
- `NttManager`, Contract to bridge out of Ethereum, see https://docs.swissborg.com/bridge/contract-addresses
- `WormholeTransceiver`, Contract to bridge in on Ethereum, see https://docs.swissborg.com/bridge/contract-addresses

### DIFF
--- a/registry/swissborg/calldata-ChsbToBorgMigrator.json
+++ b/registry/swissborg/calldata-ChsbToBorgMigrator.json
@@ -1,0 +1,393 @@
+{
+    "$schema": "../../specs/erc7730-v1.schema.json",
+    "context": {
+      "$id": "ChsbToBorgMigrator",
+      "contract": {
+        "deployments": [
+          {
+            "chainId": 1,
+            "address": "0xaA854688caAB725fe17b7D21b46fDA5AF365985a"
+          }
+        ],
+        "abi": [
+          {
+            "type": "function",
+            "name": "BORG",
+            "inputs": [],
+            "outputs": [
+              {
+                "name": "",
+                "type": "address",
+                "internalType": "contract IERC20Upgradeable",
+                "components": null,
+                "indexed": null,
+                "unit": null
+              }
+            ],
+            "stateMutability": "view",
+            "constant": null,
+            "payable": null,
+            "gas": null,
+            "signature": null
+          },
+          {
+            "type": "function",
+            "name": "CHSB",
+            "inputs": [],
+            "outputs": [
+              {
+                "name": "",
+                "type": "address",
+                "internalType": "contract IERC20Upgradeable",
+                "components": null,
+                "indexed": null,
+                "unit": null
+              }
+            ],
+            "stateMutability": "view",
+            "constant": null,
+            "payable": null,
+            "gas": null,
+            "signature": null
+          },
+          {
+            "type": "function",
+            "name": "getImplementation",
+            "inputs": [],
+            "outputs": [
+              {
+                "name": "",
+                "type": "address",
+                "internalType": "address",
+                "components": null,
+                "indexed": null,
+                "unit": null
+              }
+            ],
+            "stateMutability": "view",
+            "constant": null,
+            "payable": null,
+            "gas": null,
+            "signature": null
+          },
+          {
+            "type": "function",
+            "name": "initialize",
+            "inputs": [
+              {
+                "name": "_chsb",
+                "type": "address",
+                "internalType": "address",
+                "components": null,
+                "indexed": null,
+                "unit": null
+              },
+              {
+                "name": "_borg",
+                "type": "address",
+                "internalType": "address",
+                "components": null,
+                "indexed": null,
+                "unit": null
+              },
+              {
+                "name": "owner_",
+                "type": "address",
+                "internalType": "address",
+                "components": null,
+                "indexed": null,
+                "unit": null
+              },
+              {
+                "name": "_manager",
+                "type": "address",
+                "internalType": "address",
+                "components": null,
+                "indexed": null,
+                "unit": null
+              }
+            ],
+            "outputs": [],
+            "stateMutability": "nonpayable",
+            "constant": null,
+            "payable": null,
+            "gas": null,
+            "signature": null
+          },
+          {
+            "type": "function",
+            "name": "manager",
+            "inputs": [],
+            "outputs": [
+              {
+                "name": "",
+                "type": "address",
+                "internalType": "address",
+                "components": null,
+                "indexed": null,
+                "unit": null
+              }
+            ],
+            "stateMutability": "view",
+            "constant": null,
+            "payable": null,
+            "gas": null,
+            "signature": null
+          },
+          {
+            "type": "function",
+            "name": "migrate",
+            "inputs": [
+              {
+                "name": "_amount",
+                "type": "uint256",
+                "internalType": "uint256",
+                "components": null,
+                "indexed": null,
+                "unit": null
+              }
+            ],
+            "outputs": [],
+            "stateMutability": "nonpayable",
+            "constant": null,
+            "payable": null,
+            "gas": null,
+            "signature": null
+          },
+          {
+            "type": "function",
+            "name": "owner",
+            "inputs": [],
+            "outputs": [
+              {
+                "name": "",
+                "type": "address",
+                "internalType": "address",
+                "components": null,
+                "indexed": null,
+                "unit": null
+              }
+            ],
+            "stateMutability": "view",
+            "constant": null,
+            "payable": null,
+            "gas": null,
+            "signature": null
+          },
+          {
+            "type": "function",
+            "name": "pause",
+            "inputs": [],
+            "outputs": [],
+            "stateMutability": "nonpayable",
+            "constant": null,
+            "payable": null,
+            "gas": null,
+            "signature": null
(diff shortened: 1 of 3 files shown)


registry/swissborg/calldata-ChsbToBorgMigrator.json
```diff
@@ -0,0 +1,393 @@
...
+              "format": "tokenAmount",
+              "params": {
+                "tokenPath": "0xba9d4199faB4f26eFE3551D490E3821486f135Ba",
+                "nativeCurrencyAddress": "0xba9d4199faB4f26eFE3551D490E3821486f135Ba"
+              },
+              "path": "#._amount"
+            }
+          ],
+          "required": [
+            "amount"
```
paoun-ledger: It should be the parameter name `_amount`

registry/swissborg/calldata-ChsbToBorgMigrator.json
```diff
@@ -0,0 +1,393 @@
...
+          "$id": "migrate",
+          "intent": "Migrate CHSB to BORG",
+          "fields": [
+            {
+              "$id": "migrate",
+              "label": "CHSB Amount",
+              "format": "tokenAmount",
+              "params": {
+                "tokenPath": "0xba9d4199faB4f26eFE3551D490E3821486f135Ba",
+                "nativeCurrencyAddress": "0xba9d4199faB4f26eFE3551D490E3821486f135Ba"
```
paoun-ledger: `nativeCurrencyAddress` is probably useless in your case. It's used to display some particular addresses as the native chain currency, here ETH for chainId 1.
But here I think you meant to display CHSB, not ETH?

registry/swissborg/calldata-ChsbToBorgMigrator.json
```diff
@@ -0,0 +1,393 @@
...
+        "migrate(uint256)": {
+          "$id": "migrate",
+          "intent": "Migrate CHSB to BORG",
+          "fields": [
+            {
+              "$id": "migrate",
+              "label": "CHSB Amount",
+              "format": "tokenAmount",
+              "params": {
+                "tokenPath": "0xba9d4199faB4f26eFE3551D490E3821486f135Ba",
```
paoun-ledger: tokenPath should not contain the constant, it should contain either a path in the calldata, or a path to a constant section defined in that file. The way to do it would be to add a constant in your metadata section:

    "metadata": {
      "owner": "SwissBorg Migrator",
      "info": {
        "legalName": "SwissBorg",
        "url": "https://migration.swissborg.com"
      },
      "constants": { "CHSBaddress": "0xba9d4199faB4f26eFE3551D490E3821486f135Ba" }
    },

and then to reference that constant here in that tokenAmount:

              "$id": "migrate",
              "label": "CHSB Amount",
              "format": "tokenAmount",
              "params": {
                "tokenPath": "$.metadata.constants.CHSBaddress",

registry/swissborg/calldata-NttManager.json
```diff
@@ -0,0 +1,1455 @@
...
+        "transfer(uint256,uint16,bytes32)": {
+          "$id": "transfer",
+          "intent": "Bridge BORG",
+          "fields": [
+            {
+              "$id": "amount",
+              "label": "Amount",
+              "format": "tokenAmount",
+              "params": {
+                "tokenPath": "0x64d0f55Cd8C7133a9D7102b13987235F486F2224",
```
paoun-ledger: same as previous comment, `nativeCurrencyAddress` is probably useless here, and `tokenPath` should contain a path to a constant, defined in metadata section

registry/swissborg/calldata-NttManager.json
```diff
@@ -0,0 +1,1455 @@
...
+        "transfer(uint256,uint16,bytes32,bytes32,bool,bytes)": {
+          "$id": "transfer2",
+          "intent": "Bridge BORG",
+          "fields": [
+            {
+              "$id": "amount",
+              "label": "Amount",
+              "format": "tokenAmount",
+              "params": {
+                "tokenPath": "0xba9d4199faB4f26eFE3551D490E3821486f135Ba",
```
paoun-ledger: same comment here

registry/swissborg/calldata-ChsbToBorgMigrator.json
jnicoulaud-ledger: Little correction, to use an actual constant you need to use `token` instead of `tokenPath`:
```json
"params": {
    "token": "0xba9d4199faB4f26eFE3551D490E3821486f135Ba"
}
```
or, to reference a document constant:
```json
"params": {
    "token": "$.metadata.constants.CHSBaddress"
}
```
If using `tokenPath`, the value must be a calldata path (even if the calldata path is stored in a document constant)

antoineborg:
@paoun-ledger pushed a commit which should be answering your comments, let me know if that's not the case!

registry/swissborg/calldata-NttManager.json
```diff
@@ -0,0 +1,1457 @@
...
+        "transfer(uint256,uint16,bytes32,bytes32,bool,bytes)": {
+          "$id": "transfer2",
+          "intent": "Bridge BORG",
+          "fields": [
+            {
+              "$id": "amount",
+              "label": "Amount",
+              "format": "tokenAmount",
+              "params": {
+                "token": "$.metadata.constants.chsbAddress"
```
paoun-ledger: out of curiosity, shouldn't it be also BORG token here instead of CHSB?
Maybe both tokens have the same value, but when looking at some transactions, it seems to transfer BORG tokens: https://etherscan.io/tx/0xa7b7b8bc8f641c100d656a151c6e2cc9b6b9669b09cafe373de2039f74fea7bc

Another detail, the transactions have a non-zero value (0.000565539063465701 ETH in previous example).
Is it some kind of fee for the protocol? It may be safer to display it on the device to protect the user, in the case that value is manipulated by an attacker on software wallet side.
The syntax to reference the value within the transaction would look like this: ` "path": "@.value",`

    {
      "$id": "...",
      "label": "...",
      "path": "@.value",
      "format": "amount"
    },

paoun-ledger:
> @paoun-ledger I pushed a commit which should be answering your comments, let me know if that's not the case!

thanks @antoineborg now the syntax looks good ! I have one question on the formatter itself

registry/swissborg/calldata-NttManager.json
antoineborg: @paoun-ledger nice catch! this one is on me, mixed up the tokens, should be fixed in the latest commits!

paoun-ledger: @antoineborg thanks, it looks almost good now, but I still think the transaction value should be displayed to the user for that selector, since it's non-zero and looks like a fee.
It could be that kind of formatter:

    {
      "$id": "deliveryPrice",
      "label": "Delivery Price",
      "path": "@.value",
      "format": "amount"
    },

Not sure about the label, or maybe it could be something like `Protocol Fee`? What do you think?

antoineborg: @paoun-ledger just added it in the latest commit, it's more of a relaying fee to get the tokens directly to your wallet when you bridge. Let me know if that works for you!

paoun-ledger: Look good to me!

registry/swissborg/calldata-ChsbToBorgMigrator.json
```diff
@@ -0,0 +1,393 @@
+{
+    "$schema": "../../specs/erc7730-v1.schema.json",
+    "context": {
+      "$id": "ChsbToBorgMigrator",
```
mohammed-ledger: Consider using a more descriptive name. This field represents the `Smart Contract` name which ChsbToBorgMigrator may not be user friendly. Thank you

registry/swissborg/calldata-NttManager.json
```diff
@@ -0,0 +1,1462 @@
+{
+    "$schema": "../../specs/erc7730-v1.schema.json",
+    "context": {
+      "$id": "NttManager",
```
mohammed-ledger: Same here

registry/swissborg/calldata-WormholeTransceiver.json
```diff
@@ -0,0 +1,943 @@
+{
+    "$schema": "../../specs/erc7730-v1.schema.json",
+    "context": {
+      "$id": "WormholeTransceiver",
```
mohammed-ledger: Same here

registry/swissborg/calldata-ChsbToBorgMigrator.json
antoineborg: updated!

registry/swissborg/calldata-NttManager.json
antoineborg: updated!

registry/swissborg/calldata-WormholeTransceiver.json
antoineborg: updated!

mohammed-ledger (approved):

PR #1362, opened 2025-07-06
Update  aave/calldata-lpv3.json with more functions and merged abi

0xAkuti:
I updated the AAVE v3 pool with more functions, the initial version of this file was created using my forked version of [python-erc7730](https://github.com/0xAkuti/python-erc7730) with AI support and then all linting errors were fixed and some issues manually edited.
The file includes the merged ABI of the proxy and implementation contracts, so the abi is included explicitly, this function includes many functions missing in the original aave v3 json file.

This is a submission for the EthGlobal Cannes hackathon.

### DIFF
--- a/registry/aave/calldata-lpv3.json
+++ b/registry/aave/calldata-lpv3.json
@@ -1,166 +1,722 @@
 {
-  "$schema": "../../specs/erc7730-v1.schema.json",
   "context": {
-    "$id": "PoolInstance",
     "contract": {
-      "abi": "https://api.etherscan.io/api?module=contract&action=getabi&address=0xef434e4573b90b6ecd4a00f4888381e4d0cc5ccd",
-      "deployments": [{ "chainId": 1, "address": "0x87870Bca3F3fD6335C3F4ce8392D69350B4fA4E2" }]
+      "deployments": [{ "chainId": 1, "address": "0x87870bca3f3fd6335c3f4ce8392d69350b4fa4e2" }],
+      "abi": [
+        {
+          "type": "function",
+          "name": "approvePositionManager",
+          "inputs": [
+            { "name": "positionManager", "type": "address", "internalType": "address" },
+            { "name": "approve", "type": "bool", "internalType": "bool" }
+          ],
+          "outputs": [],
+          "stateMutability": "nonpayable"
+        },
+        {
+          "type": "function",
+          "name": "borrow",
+          "inputs": [
+            { "name": "asset", "type": "address", "internalType": "address" },
+            { "name": "amount", "type": "uint256", "internalType": "uint256" },
+            { "name": "interestRateMode", "type": "uint256", "internalType": "uint256" },
+            { "name": "referralCode", "type": "uint16", "internalType": "uint16" },
+            { "name": "onBehalfOf", "type": "address", "internalType": "address" }
+          ],
+          "outputs": [],
+          "stateMutability": "nonpayable"
+        },
+        {
+          "type": "function",
+          "name": "configureEModeCategory",
+          "inputs": [
+            { "name": "id", "type": "uint8", "internalType": "uint8" },
+            {
+              "name": "category",
+              "type": "tuple",
+              "internalType": "struct DataTypes.EModeCategoryBaseConfiguration",
+              "components": [
+                { "name": "ltv", "type": "uint16", "internalType": "uint16" },
+                { "name": "liquidationThreshold", "type": "uint16", "internalType": "uint16" },
+                { "name": "liquidationBonus", "type": "uint16", "internalType": "uint16" },
+                { "name": "label", "type": "string", "internalType": "string" }
+              ]
+            }
+          ],
+          "outputs": [],
+          "stateMutability": "nonpayable"
+        },
+        {
+          "type": "function",
+          "name": "configureEModeCategoryBorrowableBitmap",
+          "inputs": [
+            { "name": "id", "type": "uint8", "internalType": "uint8" },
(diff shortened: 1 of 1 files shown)


registry/aave/calldata-lpv3.json
```diff
@@ -1,166 +1,722 @@
 {
-  "$schema": "../../specs/erc7730-v1.schema.json",
```
paoun-ledger, line 2: It would be better to keep the reference to the `$schema` here to be future proof

registry/aave/calldata-lpv3.json
```diff
@@ -1,166 +1,722 @@
...
+      ]
     }
   },
-  "metadata": {
-    "owner": "Aave",
-    "info": { "url": "https://aave.com", "legalName": "Aave DAO", "deploymentDate": "2024-10-09T21:46:47Z" },
-    "enums": { "interestRateMode": { "1": "stable", "2": "variable" } },
-    "constants": { "max": "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff" }
-  },
+  "metadata": {},
```
paoun-ledger, line 406: Why remove the metadata? Those informations will be displayed on the device as transaction metadata, to inform the user about the contract they're interacting with

registry/aave/calldata-lpv3.json
```diff
@@ -1,166 +1,722 @@
...
   "display": {
     "formats": {
-      "repay(address asset, uint256 amount, uint256 interestRateMode, address onBehalfOf)": {
-        "$id": "repay",
-        "intent": "Repay loan",
+      "approvePositionManager(address,bool)": {
```
paoun-ledger, line 409: In all those functions, it's mandatory to add an `intent` (as what was done for `repay` selector before your PR). This will be displayed as the intent of the transaction, while all fields bellow will describe the selector parameters so both are mandatory

registry/aave/calldata-lpv3.json
```diff
@@ -1,166 +1,722 @@
...
           }
-        ],
-        "required": ["amount", "onBehalfOf", "interestRateMode"],
-        "excluded": ["referralCode"]
+        ]
       },
-      "deposit(address asset, uint256 amount, address onBehalfOf, uint16 referralCode)": {
-        "$id": "deposit",
-        "intent": "Supply",
+      "repay(address,uint256,uint256,address)": {
```
paoun-ledger, line 573: For that `repay` selector, you format many fields as raw while there are formaters available to clear sign them, you can look at what were those fields values in the previous version of this file:

          {
            "path": "amount",
            "format": "tokenAmount",
            "label": "Amount to repay",
            "params": { "tokenPath": "asset", "threshold": "$.metadata.constants.max", "message": "All" }
          },
          {
            "path": "interestRateMode",
            "format": "enum",
            "label": "Interest rate mode",
            "params": { "$ref": "$.metadata.enums.interestRateMode" }
          },
          {
            "path": "onBehalfOf",
            "format": "addressName",
            "label": "For debt holder",
            "params": { "types": ["eoa"], "sources": ["local", "ens"] }
          }


* Instead of displaying `Asset` as an address and `Amount` as a raw value (without unit nor ticker), display both as a single line using `tokenAmount` formatter (ie. 5 USDT)
* `interestRateMode` as a displayable enum instead of raw value `1` or `2`
* `onBehalfOf` is probably an `eoa`, I don't think there is a use case where it's a token, a collection, or a contract, or I am mistaken?

Those comments apply for all the new selectors in that file

registry/aave/calldata-lpv3.json
```diff
@@ -1,166 +1,722 @@
...
           },
+          { "label": "Approve", "format": "raw", "path": "#.approve" }
+        ]
+      },
+      "borrow(address,uint256,uint256,uint16,address)": {
+        "fields": [
+          { "label": "Asset", "format": "addressName", "params": { "types": ["token"] }, "path": "#.asset" },
+          { "label": "Amount", "format": "amount", "path": "#.amount" },
+          { "label": "Interest Rate Mode", "format": "raw", "path": "#.interestRateMode" },
+          { "label": "Referral Code", "format": "raw", "path": "#.referralCode" },
```
paoun-ledger, line 425: fields that are not related to the security of the transaction, and that are now easy to review for the end user should not be displayed, to avoid making the reviewing process too complicated for non technical users. I think the referralCode don't need to be part of those displayed fields.
If you agree, you can remove that line, and add it to the excluded fields just bellow:

        "fields": [
          ...
        ],
        "excluded": ["#.referralCode"]

(same comment on other selectors)

registry/aave/calldata-lpv3.json
```diff
@@ -1,166 +1,722 @@
...
             "format": "addressName",
-            "label": "Collateral recipient",
-            "params": { "types": ["eoa"], "sources": ["local", "ens"] }
+            "params": { "types": ["wallet", "eoa", "contract", "token", "collection"] },
+            "path": "#.onBehalfOf"
+          },
+          { "label": "Deadline", "format": "date", "params": { "encoding": "timestamp" }, "path": "#.deadline" },
+          { "label": "Permit V", "format": "raw", "path": "#.permitV" },
+          { "label": "Permit R", "format": "raw", "path": "#.permitR" },
+          { "label": "Permit S", "format": "raw", "path": "#.permitS" }
```
paoun-ledger, line 607: Same as my previous comment on `referalCode`, I think RSV permit signature is not human readable and don't need to be reviewed manually by the user (the permit itself was reviewed anyway). Therefore those 3 params should probably be excluded

(same on other selectors)

PR #1377, opened 2025-07-07
feat: add embedded calldata examples

paoun-ledger:


specs/erc-7730.md
```diff
@@ -1048,13 +1115,14 @@ Formats usable for bytes
...
+| `amount`                 | Optional associated amount in native currency, if the calldata can be associated with a container value. If set, the relevant ERC-7730 file matching this embedded calldata can refers this value using `@.value` container path.                                                                                                                                                                                            |
+| *Examples*               |                                                                                                                                                                                                                                                                                                                                                                                                                              |
```
paoun-ledger, line 1128: is there a use case where the calldata is for another chainId than its parent tx? If that's something possible we may want to add an optional chainId parameter

registry/safe/calldata-GnosisSafe.json
```diff
@@ -0,0 +1,111 @@
...
+      "deployments": [
+        { "chainId": 1, "address": "0x3E5c63644E683549055b9Be8653de26E0B4CD36E" },
+        { "chainId": 1, "address": "0xd9Db270c1B5E3Bd161E8c8503c55cEABeE709552" }
+      ]
+    }
+  },
+  "metadata": {
+    "owner": "Safe",
+    "info": { "url": "https://app.safe.global/welcome", "legalName": "Safe{Wallet}", "deploymentDate": "2021-05-25T03:53:22Z" },
+    "enums": { "operation": { "0": "Call", "1": "Delegate Call" } },
```
paoun-ledger, line 6: note: `Delegate Call` is considered risky. Should we have a more scary label? Or should we keep the warning for web3checks?

registry/safe/calldata-GnosisSafe.json
```diff
@@ -0,0 +1,111 @@
...
+            "format": "raw"
+          },
+          {
+            "path": "gasPrice",
+            "label": "Gas price",
+            "format": "tokenAmount",
+            "params": {
+              "tokenPath": "gasToken",
+              "nativeCurrencyAddress": "$.metadata.constants.addressAsNull"
+            },
```
paoun-ledger: it's a pity having to display those gas price and gas amount, ideally they should always be 0 for our use case... But if set to a non zero value it may be used to transfer funds so it's probably better to display everything

registry/safe/calldata-GnosisSafe.json
```diff
@@ -0,0 +1,111 @@
...
+            "path": "operation",
+            "label": "Operation",
+            "format": "enum",
+            "params": { "$ref": "$.metadata.enums.operation" }
+          },
+          {
+            "path": "baseGas",
+            "label": "Gas amount",
+            "format": "raw"
+          },
```
paoun-ledger, line 92: it's a pity having to display those gas price and gas amount, ideally they should always be 0 for our use case... But if set to a non zero value it may be used to transfer funds so it's probably better to display everything

registry/safe/calldata-GnosisSafe.json
```diff
@@ -0,0 +1,112 @@
...
+        "intent": "Execute transaction",
+        "fields": [
+          {
+            "path": "data",
+            "label": "Transaction",
+            "format": "calldata",
+            "params": {
+              "calleePath": "to",
+              "amount": "value",
+              "spender": "@.to"
```
paoun-ledger: is it correct? If the generic parser has to display the spender, should it be the Safe address?

registry/safe/calldata-GnosisSafe.json
```diff
@@ -0,0 +1,112 @@
...
+            "path": "paymentReceiver",
+            "label": "Payment receiver",
+            "format": "addressName",
+            "params": { "types": ["eoa", "wallet", "contract"] }
+          },
+          {
+            "path": "data",
+            "label": "Module setup",
+            "format": "calldata",
+            "params": { "calleePath": "to" }
```
paoun-ledger, line 51: will be empty if there is no module to setup. Therefore the device should ignore the field if calldata payload is empty (instead of failing the tx)

registry/safe/calldata-GnosisSafe.json
```diff
@@ -0,0 +1,112 @@
+{
+  "$schema": "../../specs/erc7730-v1.schema.json",
+  "context": {
+    "$id": "GnosisSafe",
+    "contract": {
+      "abi": "https://api.etherscan.io/api?module=contract&action=getabi&address=0x3E5c63644E683549055b9Be8653de26E0B4CD36E",
+      "deployments": [
+        { "chainId": 1, "address": "0x3E5c63644E683549055b9Be8653de26E0B4CD36E" },
+        { "chainId": 1, "address": "0xd9Db270c1B5E3Bd161E8c8503c55cEABeE709552" }
```
paoun-ledger: is it supported, having several deployments on the same chain?

registry/safe/calldata-GnosisSafe.json
pollastri-pierre: Not too scary though because we need it to batch transactions.

registry/safe/calldata-GnosisSafe.json
pollastri-pierre: 🤔 Maybe I misunderstood but `to` is not the Safe address. It's the recipient of the funds.

pollastri-pierre: Oh the @.to you mean. Yep seems good.

pollastri-pierre (approved):

registry/safe/calldata-GnosisSafe.json
paoun-ledger: ok then let's only be factual here, and let TX checks to the risk analysis

specs/erc-7730.md
fsamier: https://github.com/LedgerHQ/app-ethereum/blob/feat/apa/nested_gcs/doc/tlv_structs.md#param_calldata it seems to be the case in TLV specification

lcastillo-ledger (approved):

PR #1589, opened 2025-09-21
Add clear signing support for Serenita's public ETH staking vault

eth2353:
Adds clear signing support for Serenita's public ETH staking DApp deployed at [app.serenita.io](https://app.serenita.io) backed by the smart contract at [0xb36fc5e542cb4fc562a624912f55da2758998113](https://etherscan.io/address/0xb36fc5e542cb4fc562a624912f55da2758998113) .

This is a standard [StakeWise V3](https://github.com/stakewise/v3-core) smart contract, audits available in the repo.

### DIFF
--- a/registry/serenita/calldata-EthVault.json
+++ b/registry/serenita/calldata-EthVault.json
@@ -1,0 +1,240 @@
+{
+  "$schema": null,
+  "context": {
+    "$id": "ETH Staking Vault",
+    "contract": {
+      "deployments": [{ "chainId": 1, "address": "0xb36fc5e542cb4fc562a624912f55da2758998113" }],
+      "abi": [
+        {
+          "type": "function",
+          "name": "claimExitedAssets",
+          "inputs": [
+            {
+              "name": "positionTicket",
+              "type": "uint256",
+              "internalType": "uint256",
+              "components": null,
+              "indexed": null,
+              "unit": null
+            },
+            { "name": "timestamp", "type": "uint256", "internalType": "uint256", "components": null, "indexed": null, "unit": null },
+            {
+              "name": "exitQueueIndex",
+              "type": "uint256",
+              "internalType": "uint256",
+              "components": null,
+              "indexed": null,
+              "unit": null
+            }
+          ],
+          "outputs": [],
+          "stateMutability": "nonpayable",
+          "constant": null,
+          "payable": null,
+          "gas": null,
+          "signature": null
+        },
+        {
+          "type": "function",
+          "name": "deposit",
+          "inputs": [
+            { "name": "receiver", "type": "address", "internalType": "address", "components": null, "indexed": null, "unit": null },
+            { "name": "referrer", "type": "address", "internalType": "address", "components": null, "indexed": null, "unit": null }
+          ],
+          "outputs": [{ "name": "shares", "type": "uint256", "internalType": "uint256", "components": null, "indexed": null, "unit": null }],
+          "stateMutability": "payable",
+          "constant": null,
+          "payable": null,
+          "gas": null,
+          "signature": null
+        },
+        {
+          "type": "function",
+          "name": "enterExitQueue",
+          "inputs": [
+            { "name": "shares", "type": "uint256", "internalType": "uint256", "components": null, "indexed": null, "unit": null },
+            { "name": "receiver", "type": "address", "internalType": "address", "components": null, "indexed": null, "unit": null }
+          ],
+          "outputs": [
+            {
+              "name": "positionTicket",
+              "type": "uint256",
+              "internalType": "uint256",
+              "components": null,
+              "indexed": null,
+              "unit": null
+            }
+          ],
+          "stateMutability": "nonpayable",
+          "constant": null,
+          "payable": null,
+          "gas": null,
+          "signature": null
+        },
+        {
+          "type": "function",
+          "name": "multicall",
+          "inputs": [{ "name": "data", "type": "bytes[]", "internalType": "bytes[]", "components": null, "indexed": null, "unit": null }],
+          "outputs": [{ "name": "results", "type": "bytes[]", "internalType": "bytes[]", "components": null, "indexed": null, "unit": null }],
+          "stateMutability": "nonpayable",
+          "constant": null,
+          "payable": null,
+          "gas": null,
+          "signature": null
+        },
+        {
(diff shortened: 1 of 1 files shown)


registry/serenita/calldata-EthVault.json
```diff
@@ -0,0 +1,212 @@
...
+  },
+  "metadata": { "owner": "Serenita", "info": { "legalName": "Serenita B.V.", "url": "https://serenita.io/" } },
+  "display": {
+    "formats": {
+      "claimExitedAssets(uint256,uint256,uint256)": {
+        "intent": "claim exited assets",
+        "fields": [
+          { "label": "Position Ticket", "format": "raw", "path": "#.positionTicket" },
+          { "label": "Timestamp", "format": "date", "params": { "encoding": "timestamp" }, "path": "#.timestamp" },
+          { "label": "Exit Queue Index", "format": "raw", "path": "#.exitQueueIndex" }
```
paoun-ledger: Usually we want to only clear sign data that is human readable, and that can be reviewed by the end user. If I'm not mistaken, `Exit Queue Index` and `Position Ticket` are rather internal data of that smart contract, and cannot really be verified by the end user?
If that's the case, it may be better to exclude those fields

registry/serenita/calldata-EthVault.json
```diff
@@ -0,0 +1,212 @@
...
+        ],
+        "required": ["#.positionTicket", "#.timestamp", "#.exitQueueIndex"]
+      },
+      "deposit(address,address)": {
+        "intent": "deposit and stake ETH",
+        "fields": [
+          {
```
paoun-ledger, line 154: You should display the amount sent by the user. It's a missing feature from the generator UI, but actually the TX value can be displayed using the path `@.value`, fore example:

        { 
          "label": "Amount to deposit",
          "format": "amount",
          "path": "@.value" 
        }

registry/serenita/calldata-EthVault.json
```diff
@@ -0,0 +1,212 @@
...
+            "path": "#.receiver"
+          }
+        ],
+        "required": ["#.receiver"],
+        "excluded": ["#.referrer"]
+      },
+      "enterExitQueue(uint256,address)": {
+        "intent": "enter exit queue",
+        "fields": [
+          { "label": "Shares to lock", "format": "raw", "path": "#.shares" },
```
paoun-ledger, line 168: Out of curiosity, is this an amount of tokens? For instance, will it always be an amount of `osETH`, or is it not related?
Because if it was the case, you could have a more human readable display using the `tokenAmount` formatter:

        { 
          "label": "Shares to lock",
          "format": "tokenAmount",
          "path": "#.shares" 
          "params": { "token": "0xf1c9acdc66974dfb6decb12aa385b9cd01190e38" },
        }

registry/serenita/calldata-EthVault.json
```diff
@@ -0,0 +1,212 @@
...
+      "updateState((bytes32,int160,uint160,bytes32[]))": {
+        "intent": "update vault's internal state",
+        "fields": [],
+        "required": [],
+        "excluded": ["#.harvestParams.rewardsRoot", "#.harvestParams.reward", "#.harvestParams.unlockedMevReward", "#.harvestParams.proof.[]"]
+      },
+      "updateStateAndDeposit(address,address,(bytes32,int160,uint160,bytes32[]))": {
+        "intent": "deposit and stake ETH",
+        "fields": [
+          {
```
paoun-ledger, line 192: same as for `deposit`, the TX value should be displayed to the user

registry/serenita/calldata-EthVault.json
eth2353: Thanks for the review @paoun-ledger !

It makes sense to exclude those fields here, they are all only internally relevant.

I removed the timestamp too since it doesn't represent the current timestamp, but the timestamp at which the exit was initiated previously ( when calling `enterExitQueue` ).

registry/serenita/calldata-EthVault.json
eth2353: Seems like a very useful feature. Addressed in bcd6cfb89c4cb06543e9d47728c81e2e0878441d

registry/serenita/calldata-EthVault.json
eth2353: Addressed in bcd6cfb89c4cb06543e9d47728c81e2e0878441d

registry/serenita/calldata-EthVault.json
eth2353: No, these shares don't represent osETH tokens.

Our staking vault does not emit tokens, instead it internally tracks the amount of shares that belongs to each depositor which gives them a corresponding claim on ETH assets inside the vault. Kind of like an ERC20 contract, but with internal shares instead of actual tokens.

This tracking mechanism is necessary due to the staking yield - the corresponding claim on ETH changes constantly, whereas the depositor's number of shares remains the same.

The vault has a conversion function ([convertToAssets](https://etherscan.io/address/0xb36fc5e542cb4fc562a624912f55da2758998113#readProxyContract#F5)) through which you can convert the internal shares to ETH.

Since this amount of shares is not directly understandable to users, perhaps it may be better to exclude this field too, leaving the "Receiver of exited assets" clear signed?

paoun-ledger: ok thanks, very clear.
I think we can still keep it even if not easily verifiable

registry/serenita/calldata-EthVault.json
```diff
@@ -0,0 +1,211 @@
...
+        ],
+        "required": ["#.receiver"],
+        "excluded": ["#.referrer"]
+      },
+      "enterExitQueue(uint256,address)": {
+        "intent": "enter exit queue",
+        "fields": [
+          { "label": "Shares to lock", "format": "raw", "path": "#.shares" },
+          {
+            "label": "Receiver of exited assets",
```
paoun-ledger: one last detail, the Ledger device will crop labels bigger than 20 characters due to the small screen. Maybe you could find another wording for this one, or even just keep `Receiver`?

eth2353: Of course, changed to just "Receiver".

paoun-ledger (approved):

mohammed-ledger (approved):

PR #1599, opened 2025-09-23
feat: added calldata for contracts used by p2p dapp

rafal2228:
Adding SC calldata used by https://eth.p2p.org

Note there will be a separate dedicated deployment of this app for Ledger Live purposes

### DIFF
--- a/registry/consensus-specs/calldata-DepositContract.json
+++ b/registry/consensus-specs/calldata-DepositContract.json
@@ -148,7 +148,7 @@
         "fields": [
           {
             "$id": null,
-            "label": "Public key",
+            "label": "Validator public key",
             "format": "raw",
             "params": null,
             "path": "#.pubkey",


registry/consensus-specs/calldata-DepositContract.json
```diff
@@ -0,0 +1,187 @@
...
+      "factory": null
+    }
+  },
+  "metadata": {
+    "owner": "Ethereum",
+    "info": {
+      "legalName": "Ethereum Foundation",
+      "url": "https://ethereum.foundation"
+    }
+  },
```
rafal2228, line 141: Not sure about this but wanted to indicate that this is a native DepositContract and not something we (as p2p) have created/deployed

registry/p2p/calldata-EigenPodManager.json
```diff
@@ -0,0 +1,709 @@
...
+  "metadata": {
+    "owner": "eigenlayer",
+    "info": {
+      "legalName": "Eigen Labs, Inc.",
+      "url": "https://www.eigencloud.xyz/"
+    }
+  },
+  "display": {
+    "formats": {
+      "createPod()": {
```
rafal2228, line 699: Not sure about this - the [metadata UI](https://get-clear-signed.ledger.com/) doesn't allow to specify display for function calls that don't have any parameters, so I had to add this manually - but we wanted to add intent for the user to know what transaction they will be sending

paoun-ledger: This is indeed possible and missing from the UI

registry/p2p/calldata-NativeTokenVault.json
```diff
@@ -0,0 +1,1014 @@
...
+    }
+  },
+  "display": {
+    "formats": {
+      "deposit(address,address)": {
+        "$id": null,
+        "intent": "Deposit ETH with p2p.org pooled staking",
+        "screens": null,
+        "fields": [
+          {
```
paoun-ledger, line 848: There is an important missing field which is the value to deposit, but it's not your fault because the UI currently don't allow adding data from the TX itself in the fields.
You ad add it like that:

        { 
          "label": "Amount to deposit",
          "format": "amount",
          "path": "@.value" 
        }

registry/p2p/calldata-NativeTokenVault.json
```diff
@@ -0,0 +1,1014 @@
...
+    "info": {
+      "legalName": "P2P Staking",
+      "url": "https://www.p2p.org/"
+    }
+  },
+  "display": {
+    "formats": {
+      "deposit(address,address)": {
+        "$id": null,
+        "intent": "Deposit ETH with p2p.org pooled staking",
```
paoun-ledger: The intent should usually be small since it have to be display on the hardware wallet screen which is also quite small... Maybe you could rephrase it, or remove `with p2p.org` that will be displayed anyway in the metadata?

registry/p2p/calldata-NativeTokenVault.json
```diff
@@ -0,0 +1,1014 @@
...
+        ],
+        "required": ["#.receiver"],
+        "excluded": null
+      },
+      "updateStateAndDeposit(address,address,(bytes32,int160,uint160,bytes32[]))": {
+        "$id": null,
+        "intent": "Deposit ETH with p2p.org pooled staking",
+        "screens": null,
+        "fields": [
+          {
```
paoun-ledger, line 887: since it's a deposit, I think the `@.value` field should also be added here?

registry/p2p/calldata-NativeTokenVault.json
```diff
@@ -0,0 +1,1014 @@
...
+            "path": "#.referrer",
+            "value": null
+          },
+          {
+            "path": "#.harvestParams",
+            "value": null,
+            "fields": [
+              {
+                "$id": null,
+                "label": "Rewards merkle root",
```
paoun-ledger: usually we want to only clear sign data that is verifiable by a human, but not data for internal use of the smart contract.
Here I think the merkle proof and merkle root are not readable and should not be displayed, so you should probably remove those fields.
As for the paths `reward` and `unlockedMevReward`, is it something that should be reviewed by the user for its security, or is it internal data for the contract?

registry/p2p/calldata-NativeTokenVault.json
```diff
@@ -0,0 +1,1014 @@
...
+        "excluded": null
+      },
+      "claimExitedAssets(uint256,uint256,uint256)": {
+        "$id": null,
+        "intent": "Claim exited assets from pooled staking at p2p.org",
+        "screens": null,
+        "fields": [
+          {
+            "$id": null,
+            "label": "ID of position to claim",
```
paoun-ledger: same comment here, I don't know if those IDs, timestamp, and exit queue index really have to be verified manually by the user for its security? If I'm not mistaken, the important verifications are rather done when calling `enterExitQueue`, no?

registry/p2p/calldata-P2pOrgUnlimitedEthDepositor.json
```diff
@@ -0,0 +1,707 @@
...
+    }
+  },
+  "display": {
+    "formats": {
+      "addEth(bytes32,uint96,address,(uint96,address),(uint96,address),bytes)": {
+        "$id": null,
+        "intent": "Stake ETH with p2p.org",
+        "screens": null,
+        "fields": [
+          {
```
paoun-ledger, line 602: same as in my previous comment, the TX value `@.value` should be part of the fields

registry/p2p/calldata-P2pOrgUnlimitedEthDepositor.json
```diff
@@ -0,0 +1,707 @@
...
+                  "sources": null
+                },
+                "path": "recipient",
+                "value": null
+              }
+            ]
+          },
+          {
+            "$id": null,
+            "label": "Extra Data",
```
paoun-ledger: If I'm not mistaken, those `Extra Data` are not human readable, or verifiable by the end user? If that's the case, that field should be removed

registry/p2p/calldata-P2pSsvProxyFactory.json
```diff
@@ -0,0 +1,2286 @@
...
+                  "sources": null
+                },
+                "path": "recipient",
+                "value": null
+              }
+            ]
+          },
+          {
+            "$id": null,
+            "label": "Extra Data",
```
paoun-ledger: Same comments as in the previous file, `Extra Data` seems unnecessary, and `@.value` is not displayed

registry/consensus-specs/calldata-DepositContract.json
paoun-ledger: I would suggest to put those files not related directly to P2P smart contract in another directory

registry/p2p/calldata-NativeTokenVault.json
paoun-ledger: actually the timestamp make sense and is verifiable by the user, but still not sure about position ID and exit queue index

registry/consensus-specs/calldata-DepositContract.json
rafal2228: Moved it to `consensus_specs` to match the name of the repo on eth side - https://github.com/ethereum/consensus-specs/tree/master/solidity_deposit_contract

Unless you have some other suggestion?

[chore: moved consensus deposit contract to dedicated folder](https://github.com/LedgerHQ/clear-signing-erc7730-registry/pull/1599/commits/aab6476c96cd7e4e504f7aa2c9a0b60ae67fe06c)

registry/p2p/calldata-NativeTokenVault.json
rafal2228: [fix: added value of transaction to stake transactions](https://github.com/LedgerHQ/clear-signing-erc7730-registry/pull/1599/commits/413dbbc0beb115734557260567acb4e4c829ed1a)

registry/p2p/calldata-NativeTokenVault.json
rafal2228: [fix: added value of transaction to stake transactions](https://github.com/LedgerHQ/clear-signing-erc7730-registry/pull/1599/commits/413dbbc0beb115734557260567acb4e4c829ed1a)

registry/p2p/calldata-P2pOrgUnlimitedEthDepositor.json
rafal2228: [fix: added value of transaction to stake transactions](https://github.com/LedgerHQ/clear-signing-erc7730-registry/pull/1599/commits/413dbbc0beb115734557260567acb4e4c829ed1a)

registry/p2p/calldata-NativeTokenVault.json
rafal2228: [chore: shorten intent of pooled staking deposit](https://github.com/LedgerHQ/clear-signing-erc7730-registry/pull/1599/commits/3c60ee35b5626cffd3b90441412ecfff3c8f480a)

registry/p2p/calldata-P2pOrgUnlimitedEthDepositor.json
rafal2228: [chore: hidden secondary parameters from clear signing](https://github.com/LedgerHQ/clear-signing-erc7730-registry/pull/1599/commits/4167d6b2ec2ea02b10567f3dad1fc85d245e2748)

registry/p2p/calldata-P2pSsvProxyFactory.json
rafal2228: [chore: hidden secondary parameters from clear signing](https://github.com/LedgerHQ/clear-signing-erc7730-registry/pull/1599/commits/4167d6b2ec2ea02b10567f3dad1fc85d245e2748)

registry/p2p/calldata-NativeTokenVault.json
rafal2228: [chore: hidden secondary parameters from clear signing](https://github.com/LedgerHQ/clear-signing-erc7730-registry/pull/1599/commits/4167d6b2ec2ea02b10567f3dad1fc85d245e2748)

registry/p2p/calldata-NativeTokenVault.json
rafal2228: [chore: hidden secondary parameters from clear signing](https://github.com/LedgerHQ/clear-signing-erc7730-registry/pull/1599/commits/4167d6b2ec2ea02b10567f3dad1fc85d245e2748)

registry/p2p/calldata-EigenPodManager.json
```diff
@@ -0,0 +1,709 @@
...
+          "gas": null,
+          "signature": null
+        }
+      ],
+      "addressMatcher": null,
+      "factory": null
+    }
+  },
+  "metadata": {
+    "owner": "eigenlayer",
```
paoun-ledger: small detail but since it will be displayed on the device, you could add uppercase as you did in the intent? `EigenLayer`

rafal2228: [chore: uppercase eigenlayer name](https://github.com/LedgerHQ/clear-signing-erc7730-registry/pull/1599/commits/d991dee0d80a405fff757eb04759e500f568d18a)

registry/p2p/calldata-NativeTokenVault.json
paoun-ledger: actually, having those paths in "excluded" list is not enough, to hide them you must also remove them from that field list as well. For that selector, it may result in:

        "fields": [
          {
            "$id": null,
            "label": "Exit initiated at",
            "format": "date",
            "params": {
              "encoding": "timestamp"
            },
            "path": "#.timestamp",
            "value": null
          }
        ],
        "required": ["#.timestamp"],
        "excluded": ["#.positionTicket", "#.exitQueueIndex"]

Same comment for other selectors hidden fields

registry/p2p/calldata-NativeTokenVault.json
```diff
@@ -0,0 +1,1029 @@
...
+      },
+      "enterExitQueue(uint256,address)": {
+        "$id": null,
+        "intent": "Initiate exit of funds",
+        "screens": null,
+        "fields": [
+          {
+            "$id": null,
+            "label": "Amount of shares to exit",
+            "format": "amount",
```
paoun-ledger, line 954: `"amount"` formatter is to format an amount in native currency such as ETH. From what I understand, here it's not ETH nor an erc20, it's an amount of shares that cannot be formatted correctly by the device, do you confirm? If that's the case, then you should rather use the `"raw"` formatter

registry/p2p/calldata-P2pOrgUnlimitedEthDepositor.json
paoun-ledger: same as previous comment, that field should be removed, not only put in `excluded` list

registry/p2p/calldata-P2pOrgUnlimitedEthDepositor.json
```diff
@@ -0,0 +1,720 @@
...
+  "display": {
+    "formats": {
+      "addEth(bytes32,uint96,address,(uint96,address),(uint96,address),bytes)": {
+        "$id": null,
+        "intent": "Stake ETH with p2p.org",
+        "screens": null,
+        "fields": [
+          {
+            "$id": null,
+            "label": "Validator withdrawal credentials",
```
paoun-ledger: Ledger devices screen is small, for that reason labels bigger than 20 characters will be truncated...
So it would be better to rephrase this one to fit in 20 characters if possible.
Same comment for several labels below: eth per validator, percentages of reward, percentage of referrer

paoun-ledger: same comment apply to P2pSsvProxyFactory

registry/p2p/calldata-NativeTokenVault.json
rafal2228: Shares are in the same format as ETH currency (which value they kinda represent anyway). I think amount is good formatter here - the value provided here will be in wei

registry/p2p/calldata-NativeTokenVault.json
rafal2228: [chore: removed hidden parameters from fields property](https://github.com/LedgerHQ/clear-signing-erc7730-registry/pull/1599/commits/719f6a1b6979d3998b3511bb73ad275ba6c30a8b)

registry/p2p/calldata-P2pOrgUnlimitedEthDepositor.json
rafal2228: [chore: removed hidden parameters from fields property](https://github.com/LedgerHQ/clear-signing-erc7730-registry/pull/1599/commits/719f6a1b6979d3998b3511bb73ad275ba6c30a8b)

registry/p2p/calldata-P2pOrgUnlimitedEthDepositor.json
rafal2228: [chore: shorten labels to fit into 20 characters](https://github.com/LedgerHQ/clear-signing-erc7730-registry/pull/1599/commits/2db6dd721c7aa8640ec9e956312b6396bc16d18e) & [chore: shorten intents to fit into 20 characters](https://github.com/LedgerHQ/clear-signing-erc7730-registry/pull/1599/commits/65bb305953477ba8b839de8cfba801890f4a02b0)

registry/consensus-specs/calldata-DepositContract.json
```diff
@@ -0,0 +1,179 @@
...
+            "params": null,
+            "path": "#.pubkey",
+            "value": null
+          },
+          {
+            "$id": null,
+            "label": "Withdraw credentials",
+            "format": "raw",
+            "params": null,
+            "path": "#.withdrawal_credentials",
```
paoun-ledger, line 162: Hello @rafal2228 ,
Sorry for the late an iterative review but that comment should be the last one :sweat_smile: 

Actually the withdrawal credentials will not be easy to understand by the end user as it's not formatted.
Also if it's a BLS (0x00) credential, it's almost impossible to clear sign on a ledger without custom commands since the user would require to do BLS key derivation and hashing. So I would advice to blind sign those kind of 0x00 legacy credentials, and have a better formatting for other kind of credentials.

It could be done using an enum. You could declare it this way in the metadata section (feel free to rephrase):


        "enums": {
          "credentialsType": {
            "1": "Execution key",
            "2": "Compounding account"
          }
        }

here I voluntarily don't declare the type `0` to not clear sign it (cannot be verified by the user).
Then you can have a better formatting in the display section, using slices:



              {
                "$id": null,
                "label": "Withdraw address",
                "format": "addressName",
                "params": { "types": ["eoa", "contract"] },
                "path": "#.withdrawal_credentials.[12:]",
                "value": null
              },
              {
                "$id": null,
                "label": "Account type",
                "format": "enum",
                "params": { "$ref": "$.metadata.enums.credentialsType" },
                "path": "#.withdrawal_credentials.[:1]",
                "value": null
              },

you could do the same thing in P2P selectors that take withdrawal credentials in parameters.
What do you think about it?

registry/consensus-specs/calldata-DepositContract.json
```diff
@@ -0,0 +1,179 @@
...
+  "display": {
+    "formats": {
+      "deposit(bytes,bytes,bytes,bytes32)": {
+        "$id": null,
+        "intent": "Stake ETH",
+        "screens": null,
+        "fields": [
+          {
+            "$id": null,
+            "label": "Public key",
```
paoun-ledger: Detail but `Validator public key` would fit here

registry/consensus-specs/calldata-DepositContract.json
rafal2228: > Actually the withdrawal credentials will not be easy to understand by the end user as it's not formatted.

💯 agree and in case of top up (which is the only use case for direct SC call in p2p app), this field is not even taken into consideration - https://eth2book.info/latest/part2/deposits-withdrawals/staking/#top-up-deposits - as for top ups the `withdrawal_credentials` and `signature`  are ignored

> Also if it's a BLS (0x00) credential

To be honest I doubt anyone would try to create such validators these days - IMO ledger could even consider actively discouraging that.

From p2p perspective we ideally would hide `withdrawal_credentials` - since they are not need for top ups we set both `withdrawal_credentials` and `signature` to zeros - here's an example of such transaction https://hoodi.etherscan.io/tx/0x106244ea7f2e31ba43df448c7412c24e2ce7d9b6aea7ae03b0d4d9b7b6f397d0

I think your approach to parse the WC for users is great, but it wouldn't be fully compatible with top ups. I believe we should go into one of three directions:
- show data in `raw` format (current situation, IMO a good enough middle ground) - that way users creating new validators will see necessary information and users doing the top ups could simply ignore that information
- hide this value completely - ideal scenario for top ups, but it's unsafe for people creating new validators - they wouldn't be able to confirm if `withdrawal_credentials` address is what they expect it to be
- provide a way on ledger side to indicate if the transaction is meant as top up of existing validator (and ignore `withdrawal_credentials` in such case) or as new deposit. There's currently no way to do that and I'm not sure how ledger could guarantee safety for users either way in such case

tl;dr - given the way `deposit` function is being used right now, I don't think we can do much better than showing `raw` value for `withdrawal_credentials`. I would be happy to the changes you mentioned above, but topping up users will see a lot of false positive blind signing requests in such case

registry/consensus-specs/calldata-DepositContract.json
rafal2228: [chore: added Validator to pubkey label](https://github.com/LedgerHQ/clear-signing-erc7730-registry/pull/1599/commits/5890b07ec5757653c226b93a076141db5dacad7a)

registry/consensus-specs/calldata-DepositContract.json
paoun-ledger: Ok thanks, it's indeed probably better to keep displaying the raw value for all users for now since we don't yet have a way to conditionally hide fields.

mohammed-ledger (approved):

PR #1602, opened 2025-09-25
Update eip712-uniswap-V2DutchOrder.json

CremaFR:
https://ledgerhq.atlassian.net/browse/LIVE-21919 

Fix v2 dutch order token parsing by only displaying the first 1 ( what you receive). This excludes the fees taken from being displayed. 

Also added parsing for contract address of Uniswap once in the CAL

### DIFF
--- a/registry/uniswap/eip712-uniswap-V2DutchOrder.json
+++ b/registry/uniswap/eip712-uniswap-V2DutchOrder.json
@@ -78,11 +78,11 @@
             "fields": [
               {
                 "path": "endAmount",
-                "label": "Minimum amounts to receive",
+                "label": "Minimum amount to receive",
                 "format": "tokenAmount",
                 "params": { "tokenPath": "token" }
               },
-              { "path": "recipient", "label": "On Addresses", "format": "raw" }
+              { "path": "recipient", "label": "On Address", "format": "raw" }
             ]
           },
           {


registry/uniswap/eip712-uniswap-V2DutchOrder.json
```diff
@@ -66,15 +74,31 @@
             "params": { "tokenPath": "witness.baseInputToken" }
           },
           {
-            "path": "witness.baseOutputs.[]",
+            "path": "witness.baseOutputs.[0]",
             "fields": [
-              { "path": "endAmount", "label": "Minimum amounts to receive", "format": "tokenAmount", "params": { "tokenPath": "token" } },
+              {
+                "path": "endAmount",
+                "label": "Minimum amounts to receive",
```
paoun-ledger: detail but I think `amounts` was plural because of the array, it can now be singular. Same for `Addresses` just below

registry/uniswap/eip712-uniswap-V2DutchOrder.json
```diff
@@ -66,15 +74,31 @@
             "params": { "tokenPath": "witness.baseInputToken" }
           },
           {
-            "path": "witness.baseOutputs.[]",
+            "path": "witness.baseOutputs.[0]",
             "fields": [
-              { "path": "endAmount", "label": "Minimum amounts to receive", "format": "tokenAmount", "params": { "tokenPath": "token" } },
+              {
+                "path": "endAmount",
```
paoun-ledger, line 80: not related to your change but shouldn't we rather use `startAmount` since we want to display the "Minimum to receive"? Actually I don't know which one is the minimum :sweat_smile:

CremaFR: It's an auction, startAmount is the highest you can receive, and endAmount is the lowest you can receive. The more time passes the more the amount receive get lowered to incentives a solver to make the trade!

registry/uniswap/eip712-uniswap-V2DutchOrder.json
CremaFR: You're right, I renamed

paoun-ledger (approved):

PR #1616, opened 2025-09-29
feat: midas protocol

kostyamospan:


registry/midas/calldata-MinterVault.json
```diff
@@ -0,0 +1,1230 @@
...
+      "depositInstant(address,uint256,uint256,bytes32)": {
+        "intent": "instantly buy",
+        "fields": [
+          { "label": "You pay with", "format": "addressName", "params": { "types": ["token"] }, "path": "#.tokenIn" },
+          {
+            "label": "Amount",
+            "format": "unit",
+            "params": { "base": "", "decimals": "18", "prefix": "false" },
+            "path": "#.amountToken"
+          },
```
paoun-ledger, line 202: Those two fields `You pay with` and `Amount` could be merged with the formatter `tokenAmount`, to display directly an amount of tokens formatted in a human readable way:

        "fields": [
          {
            "label": "You pay with",
            "path": "#.tokenIn"
            "format": "tokenAmount",
            "params": { "tokenPath": "#.amountToken" },
          },

Same comment for all those selectors

registry/midas/calldata-MinterVault.json
```diff
@@ -0,0 +1,1230 @@
...
+          {
+            "label": "Slippage",
+            "format": "unit",
+            "params": { "base": "", "decimals": "18", "prefix": "false" },
+            "path": "#.minReceiveAmount"
+          }
```
paoun-ledger, line 208: When looking at transactions such as this one:
https://etherscan.io/tx/0x933fc88274c7e112cfaf051e0e2621acce72b546cdc240aac92adc395ce3564c
I can see that mTBILL tokens are actually received. Is that always the case in that contract?
Because if that's the case, you could also display properly the `minReceiveAmount` as an amount with the token ticker and correct magnitude, with:

        "fields": [
          {
            "label": "You receive",
            "path": "#.minReceiveAmount"
            "format": "tokenAmount",
            "params": { "token": "0xdd629e5241cbc5919847783e6c96b2de4754e438" },
          },

registry/midas/calldata-RedemptionVault.json
```diff
@@ -0,0 +1,1230 @@
...
+    "formats": {
+      "redeemInstant(address,uint256,uint256,address)": {
+        "intent": "instantly redeem",
+        "fields": [
+          { "label": "You receive", "format": "addressName", "params": { "types": ["token"] }, "path": "#.tokenOut" },
+          {
+            "label": "Amount to redeem",
+            "format": "unit",
+            "params": { "base": "", "decimals": "18", "prefix": "false" },
+            "path": "#.amountMTokenIn"
```
paoun-ledger, line 226: same comment as before, will it always be an amount of mTBILL token? If that's the case, it can be formatted accordingly with:

        "fields": [
          {
            "label": "Amount to redeem",
            "path": "#.amountMTokenIn"
            "format": "tokenAmount",
            "params": { "token": "0xdd629e5241cbc5919847783e6c96b2de4754e438" },
          },

registry/midas/calldata-RedemptionVault.json
```diff
@@ -0,0 +1,1230 @@
...
+          },
+          {
+            "label": "Slippage",
+            "format": "unit",
+            "params": { "base": "", "decimals": "18", "prefix": "false" },
+            "path": "#.minReceiveAmount"
+          },
```
paoun-ledger, line 233: Same comment as before, if that amount corresponds to the token `#.tokenOut`, then please format it with:

        "fields": [
          {
            "label": "You receive",
            "path": "#.minReceiveAmount"
            "format": "tokenAmount",
            "params": { "tokenPath": "#.tokenOut" },
          },

registry/midas/calldata-MinterVault.json
```diff
@@ -0,0 +1,1230 @@
...
+        { "chainId": 98866, "address": "0x8F38A24d064B41c990a3f47439a7a7EE713BF8Dc" },
+        { "chainId": 98866, "address": "0x23dE49C9ECb8bAaF4aBDeD123FaFbb7D5b7a0eE2" },
+        { "chainId": 98866, "address": "0xe6F0C60Fca2bd97d633a3D9D49DBEFDF19636D8c" },
+        { "chainId": 98866, "address": "0xc4E4aCA6A81794562c46DA86c20dc652bA2Af25E" },
+        { "chainId": 98866, "address": "0x71DD2570a843B0D1c74FFAb23F348193F19F18B1" },
+        { "chainId": 747474, "address": "0x175A9b122bf22ac2b193a0A775D7370D5A75268E" },
+        { "chainId": 747474, "address": "0xcb7d9A25F7b9bdd0Eee77B1cEb2894D39deBca1C" },
+        { "chainId": 1440000, "address": "0x30FBc82A72CA674AA250cd6c27BCca1Fe602f1Bb" }
+      ],
+      "abi": [
```
lcastillo-ledger, line 85: ABI section can be simplified to include only the functions being clear signed later in the "display" section

registry/midas/calldata-MinterVault.json
kostyamospan: At this moment, all the input amounts in our contracts are required to be in 18 decimals. So even if it's let's say USDC with 6 decimals, amountToken should be ^18 not ^6. Because of that, we cannot use tokenAmount formatter, unfortunately

registry/midas/calldata-MinterVault.json
kostyamospan: We have many different products and mTBILL is one of them, but each deposit vault has its own token contract, so if we hardcode the address here, we would need to have a separate calldata file for each midas product per network that we have, at this moment its gonna be around a hundred files

registry/midas/calldata-RedemptionVault.json
kostyamospan: Same as for deposit vault, we have many different products, so hardcoding of the address will lead to hundreds of calldata files

registry/midas/calldata-RedemptionVault.json
kostyamospan: same as for deposit vault, cannot use tokenAmount because of strict 18 decimals values

registry/midas/calldata-MinterVault.json
kostyamospan: fixed

registry/midas/calldata-MinterVault.json
```diff
@@ -0,0 +1,246 @@
...
+        { "chainId": 9745, "address": "0x2b690Cab819A815732544aEb422474EfDc1B0615" },
+        { "chainId": 9745, "address": "0x4Ef9fF56162bD3Cb5073FB20DbD355C59084093f" },
+        { "chainId": 16661, "address": "0x72a93168AE79F269DeB2b1892F2AFd7eaa800271" },
+        { "chainId": 23294, "address": "0xD7Fe0e91C05CAfdd26dA4B176eEc2b883795BDcC" },
+        { "chainId": 42161, "address": "0x643f73A6a3Ffc5d6C6Be7c97Cc30422763CFb1d0" },
+        { "chainId": 42161, "address": "0x01bfF1379CE9f0877141a18670d4214dFAf630bE" },
+        { "chainId": 42161, "address": "0x2c851A37eF2d607F198DDB259309Dcd2B398E8f9" },
+        { "chainId": 42161, "address": "0xb285f7699206C88A0aEC8bb004e42793de8139e0" },
+        { "chainId": 42161, "address": "0x9815FffE5600cF71342579f0f3E0Dd8ccBd496D8" },
+        { "chainId": 42793, "address": "0xd65BFeB71271A4408ff335E59eCf6c5b21A33a70" },
```
paoun-ledger: I did some testing and unfortunately our descriptors generation tooling don't support that chainId 42793, so we should remove those deployments at least for now

registry/midas/calldata-MinterVault.json
```diff
@@ -0,0 +1,246 @@
...
+        "excluded": ["#.referrerId"]
+      },
+      "depositInstant(address,uint256,uint256,bytes32)": {
+        "intent": "instantly buy",
+        "fields": [
+          { "label": "You pay with", "format": "addressName", "params": { "types": ["token"] }, "path": "#.tokenIn" },
+          {
+            "label": "Amount",
+            "format": "unit",
+            "params": { "base": "", "decimals": "18", "prefix": "false" },
```
paoun-ledger: the `"base"` param is mandatory so it should not be empty. In that case it's probably 10?

paoun-ledger: 'same in each occurrence bellow)

registry/midas/calldata-RedemptionVault.json
```diff
@@ -0,0 +1,300 @@
+{
```
paoun-ledger, line 1: the same comments apply to this file

registry/midas/calldata-MinterVault.json
kostyamospan: i was refering to this example from the [eip definition ](https://eips.ethereum.org/EIPS/eip-7730), seems that base is just a string/symbol that goes right after the formatted value and the prefix (in our case prefix is false so the base goes after the value). If its mandatory, should we set it to something like " tokens"? 

> | | Description | Value is converted to a float using decimals (value / 10^decimals) and displayed appending the corresponding unit. If prefix is true, the value is further converted to scientific representation, minimizing the significand and converting the exponent to an SI prefix added in front of the unit symbol | | Parameters | — | | base | The symbol of the base unit, an SI unit or other acceptable symbols like “%”, “bps” | | decimals | Number of decimals in integer representation, defaults to 0 | | prefix | A boolean indicating whether an SI prefix should be appended, defaults to False | | Examples | — | | 10h | Field Value = 10
> base = “h” | | 1.5d | Field Value = 15
> base = “d”
> decimals = 1 | | 36ks | Field Value = 36000
> base = “s”
> prefix = True |

registry/midas/calldata-MinterVault.json
kostyamospan: fixed

registry/midas/calldata-MinterVault.json
kostyamospan: @paoun-ledger, what value would you recommend using for our case?

paoun-ledger: Hello @kostyamospan 
Indeed you're right the base should be optional, and it may feel weird to display "tokens" in that case.
Unfortunately our internal generators make it mandatory for now...
Your PR look good to me, but I'm working on fixing that issue on our side before we merge it

paoun-ledger: actually after a little deep dive, it's also mandatory in our devices so it may take longer to deploy a fix for that issue... In the meantime, I would suggest to just put a space as the base `" "`
It will allow your PR to be merged, and I will modify it later when the parameter will be made optional in the device

kostyamospan: Thanks a lot! Just pushed the fixed version, also added few contract addresses to be in sync with our deployments

PR #1732, opened 2025-10-15
feat: add figment batch deposit contract

trentkrogers:
Adding calldata for the deposit function in the Figment ETH Depositor smart contract to enable clear signing on mainnet and hoodi.

### DIFF
--- a/registry/figment/calldata-figment-batch-deposit.json
+++ b/registry/figment/calldata-figment-batch-deposit.json
@@ -388,8 +388,11 @@
             {
               "$id": null,
               "label": "Amount to Deposit",
-              "format": "amount",
-              "params": null,
+              "format": "unit",
+              "params": {
+                "base": "ETH",
+                "decimals": "9"
+              },
               "path": "#.amounts_gwei.[]",
               "value": null
             }


registry/figment/calldata-figment-batch-deposit.json
```diff
@@ -0,0 +1,420 @@
...
+    "display": {
+      "formats": {
+        "deposit(bytes[],bytes[],bytes[],bytes32[],uint256[])": {
+          "$id": null,
+          "intent": "Stake ETH",
+          "screens": null,
+          "fields": [
+            {
+              "$id": null,
+              "label": "Pubkeys",
```
paoun-ledger: maybe `Validator public key` would be more understandable?

registry/figment/calldata-figment-batch-deposit.json
```diff
@@ -0,0 +1,420 @@
...
+              "$id": null,
+              "label": "Withdrawal Credentials",
+              "format": "raw",
+              "params": null,
+              "path": "#.withdrawal_credentials.[]",
+              "value": null
+            },
+            {
+              "$id": null,
+              "label": "Signatures",
```
paoun-ledger: I think `Signatures` and `Deposit Data Roots` are not human readable / reviewable, or I am mistaken? If that's the case, you should probably remove those 2 fields, and add them in the `excluded` list instead

registry/figment/calldata-figment-batch-deposit.json
```diff
@@ -0,0 +1,420 @@
...
+              "$id": null,
+              "label": "Deposit Data Roots",
+              "format": "raw",
+              "params": null,
+              "path": "#.deposit_data_roots.[]",
+              "value": null
+            },
+            {
+              "$id": null,
+              "label": "Amounts Gwei",
```
paoun-ledger: Here it will be displayed in ETH unit, so no need to put Gwei in the label. Maybe just `Amount to deposit` instead?

registry/figment/calldata-figment-batch-deposit.json
trentkrogers: Yup, done! https://github.com/LedgerHQ/clear-signing-erc7730-registry/pull/1732/commits/b08fd89fc4609286275eefa67a25b5e1a2e89d28

registry/figment/calldata-figment-batch-deposit.json
trentkrogers: Thanks, removed! https://github.com/LedgerHQ/clear-signing-erc7730-registry/pull/1732/commits/4e49beeb01480182946b6a7f2966d5203b4d06a1

registry/figment/calldata-figment-batch-deposit.json
trentkrogers: Got it. Renamed here: https://github.com/LedgerHQ/clear-signing-erc7730-registry/pull/1732/commits/580d5021b97a81469dc6fcb1d2ad831f6f29d3b6

registry/figment/calldata-figment-batch-deposit.json
```diff
@@ -0,0 +1,407 @@
...
+              "label": "Withdrawal Credentials",
+              "format": "raw",
+              "params": null,
+              "path": "#.withdrawal_credentials.[]",
+              "value": null
+            },
+            {
+              "$id": null,
+              "label": "Amount to Deposit",
+              "format": "amount",
```
paoun-ledger: Actually I just did some tests on device and it's not displayed correctly because the `amount` formatter expects an input in wei while here it's in gwei.
So what I would suggest is to use the `unit` formatter instead, with the correct decimals:


            {
              "label": "Amount to Deposit",
              "format": "unit",
              "params": { "base": "ETH", "decimals": 9 },
              "path": "#.amounts_gwei.[]",
            }

registry/figment/calldata-figment-batch-deposit.json
```diff
@@ -0,0 +1,407 @@
...
+              "$id": null,
+              "label": "Validator Public Key",
+              "format": "raw",
+              "params": null,
+              "path": "#.pubkeys.[]",
+              "value": null
+            },
+            {
+              "$id": null,
+              "label": "Withdrawal Credentials",
```
paoun-ledger: labels bigger than 20 characters will be truncated on Ledger devices screen, so it may be better to shorten this one, maybe with something like: `Withdraw credentials`

trentkrogers: Updated here: https://github.com/LedgerHQ/clear-signing-erc7730-registry/pull/1732/commits/a1037ca731198fe729895c6c16a367fd31a1fee0

registry/figment/calldata-figment-batch-deposit.json
trentkrogers: Thanks for testing on device! Using the `unit` formatter now: https://github.com/LedgerHQ/clear-signing-erc7730-registry/pull/1732/commits/147ce07602adffc4f57f6675a1f950eb997cbbc7

mohammed-ledger (approved):

PR #2224, opened 2026-01-05
fix(safe): Issues fixes

Maroutis:
### `registry/safe/calldata-SafeProxyFactory-1.3.0.json`
- `createProxyWithCallback(address,bytes,uint256,address)`: Issue: `required` referenced non-existent `data`; Fix: remove the `data` requirement so only `initializer` is shown.

### `registry/safe/calldata-SafeProxyFactory-1.4.1.json`
- `createProxyWithCallback(address,bytes,uint256,address)`: Issue: `required` referenced non-existent `data`; Fix: remove the `data` requirement so only `initializer` is shown.

### `registry/safe/calldata-SafeProxyFactory-1.5.0.json`
- `createProxyWithNonceL2(address,bytes,uint256)`: Issue: `required` referenced non-existent `data`; Fix: remove the `data` requirement so only `initializer` is shown.

### `registry/safe/common-Safe.json`
- `execTransaction(address,uint256,bytes,uint8,uint256,uint256,uint256,address,address,bytes)`: Issue: intent "sign multisig operation" was misleading at execution time. This function only executes logic on-chain; Fix: intent "Execute Safe Tx".

### `registry/safe/common-SafeProxyFactory.json`
- `createProxyWithNonce(address,bytes,uint256)`: Issue: `required` referenced non-existent `data`; Fix: remove the `data` requirement so only `initializer` is shown.


registry/safe/common-Safe.json
```diff
@@ -55,7 +55,7 @@
       },
       "execTransaction(address to, uint256 value, bytes data, uint8 operation, uint256 safeTxGas, uint256 baseGas, uint256 gasPrice, address gasToken, address refundReceiver, bytes signatures)": {
         "$id": "execTransaction",
-        "intent": "sign multisig operation",
+        "intent": "Execute Safe Tx",
```
paoun-ledger: I think that wording was agreed with the product team, even if we can of course rephrase it is it's misleading

Maroutis: So do you agree with new wording ?

lcastillo-ledger: Better to keep initial wording even if truncated on some devices

Maroutis: Will do

lcastillo-ledger (approved):

fsamier (approved):

lcastillo-ledger:
4 required issues fixed by migration to v2
Wording changes on execTransaction should go through product approval

PR #2281, opened 2026-01-21
feat: add musig auth message

drakoFukayu:


registry/safe/common-eip712-Safe.json
```diff
@@ -178,6 +211,24 @@
...
+        "intent": "Authenticate address book",
+        "fields": [],
+        "required": [],
+        "excluded": ["totp"]
+      },
+      "AddAddressBookEntry": {
+        "intent": "Add address book entry",
+        "fields": [
+          {
+            "path": "entries",
```
paoun-ledger: I think it will not work, `entries` is an array of tuples so you cannot just format it in raw.
You can only display fields of "primary types", and if it's within an array, you can iterate with a path like this one:
`entries.[].alias`
or
`entries.[].address`

Another limitation is that you cannot "bundle" 2 array together to iterate on both of them at the same time. It means the device will display:

- Alias1
- Alias2
- Address1
- Address2

instead of

- Alias1
- Address1
- Alias2
- Address2
- ...

We're working on a way to improve that in ERC7730 v2 but not implemented yet on the app side. So probably for the first version you may want to have only 1 element in that array?

paoun-ledger: ok my bad, I just tested it and in eip712 (contrary to transactions), the iteration is bundled 👍 
But to make it work, you should still describe each primary type independently:


      "AddAddressBookEntry": {
        "intent": "Add address book entry",
        "fields": [
          {
            "path": "entries.[].alias",
            "label": "Address book alias",
            "format": "raw"
          },
          {
            "path": "entries.[].address",
            "label": "Address book address",
            "format": "raw"
          }
        ],
        "required": ["entries"],
        "excluded": ["totp"]
      }

drakoFukayu: Thanks for testing and confirming the bundled iteration behavior! I've updated the display format to use array iteration paths as suggested:

```json
"fields": [
  {
    "path": "entries.[].alias",
    "label": "Address book alias",
    "format": "raw"
  },
  {
    "path": "entries.[].address",
    "label": "Address book address",
    "format": "raw"
  }
]
```

This should now properly display each address book entry with its alias and address bundled together.

Fixed in commit e2ea014

PR #2503, opened 2026-03-27
feat: widen v2 schema fields to accept constants and map references

lcastillo-ledger:
## Summary

- Add reusable JSON schema definitions (`stringValueOrMap`, `integerValueOrMap`, `stringOrArrayValueOrMap`, `scalarValueOrMap`) to factor constant/map reference support
- Widen parameter fields (`token`, `callee`, `selector`, `amount`, `spender`, `chainId`, `collection`, `nativeCurrencyAddress`, `threshold`, `message`, `senderAddress`) to accept map references
- Widen metadata fields (`owner`, `contractName`) and context fields (`domain.name`, `domain.version`) to accept constant and map references
- Widen display fields (`label`, `value`, `interpolatedIntent`, `intent`) to accept map references
- Tighten `mapReference` definition: enforce `required: ["map", "keyPath"]` and `additionalProperties: false`
- Allow optional `$id` on display `reference` entries (merged from #2506)
- Update spec wording for `eip712.domain`, `intent`, `label`, `value`, and reference overrides to document the new reference forms

Companion linter PR: https://github.com/LedgerHQ/python-erc7730/pull/291

## Authors

- @lcastillo-ledger — schema and spec changes for constants/map references
- @Maroutis — `$id` support on display references (#2506)

## Test plan

- [x] Schema validates existing registry descriptors (no breaking changes)
- [x] Linter updated and tested against canonical `example-maps.json` from ERC repo
- [x] Spec wording changes are minimal and backward-compatible
- [ ] Validate a descriptor using a reference override with both `$ref` and `$id`

Made with [Cursor](https://cursor.com)

PR #2506, opened 2026-03-27
fix: allow  on display references

Maroutis:
## Summary
- allow optional `$id` on `display.reference` entries in `specs/erc7730-v2.schema.json`
- document in `specs/erc-7730.md` that reference overrides may carry an internal `$id`
- fix validation for descriptors that reuse a shared definition while assigning a field-level identifier

## Test plan
- [x] Parse `specs/erc7730-v2.schema.json` with Python `json.loads`
- [ ] Validate a descriptor using a reference override with both `$ref` and `$id`

Made with [Cursor](https://cursor.com)


lcastillo-ledger:
Merged into #2503

PR #2511, opened 2026-04-04
Add DeadSwitch InheritanceVault descriptor

Lastoneparis:
## Summary

Adding ERC-7730 clear signing descriptor for **DeadSwitch** — a decentralized dead man's switch for crypto inheritance built at ETHGlobal Cannes 2026.

## What is DeadSwitch?

DeadSwitch lets crypto holders protect their assets for their heirs:
- Owner deposits ETH into a smart contract vault
- Owner designates an heir (by wallet address or ENS name)
- Owner sends a monthly heartbeat to prove they're still alive
- If heartbeats stop, Chainlink Automation triggers recovery mode
- Heir verifies identity with World ID and claims the funds
- Fully non-custodial, on-chain

## Functions Covered

The descriptor provides clear signing for all user-facing vault interactions:

| Function | Intent |
|----------|--------|
| `heartbeat()` | Send heartbeat |
| `claim(bytes32)` | Claim inheritance |
| `cancel()` | Cancel vault & withdraw |
| `updateBeneficiary(address,bytes32,string)` | Update heir |
| `updateInterval(uint256)` | Update heartbeat interval |
| `setGuardian(address)` | Set guardian |
| `pauseRecovery()` | Pause recovery mode |
| `resumeFromRecovery()` | Resume from recovery |
| `extendHeartbeat(uint256)` | Extend heartbeat deadline |

## Deployment

- **Network:** Ethereum Sepolia Testnet (chainId 11155111)
- **Contract:** `0xF957cDA1f676B9EAE65Ab99982CAa3a31A193CB7`
- **Etherscan:** https://sepolia.etherscan.io/address/0xF957cDA1f676B9EAE65Ab99982CAa3a31A193CB7
- **Live site:** https://deadswitch.online
- **Source code:** https://github.com/Lastoneparis/deadswitch

## Why Clear Signing Matters for DeadSwitch

DeadSwitch handles **inheritance** — this is long-term, high-trust infrastructure. Owners set up a vault and may not interact with it for months. When they do (monthly heartbeat, beneficiary update, cancel), showing raw hex on a Ledger device creates anxiety: *"Am I accidentally signing away my inheritance to someone?"*

Clear signing transforms this into unambiguous confirmations:
- "Send heartbeat" — the owner knows exactly what they're signing
- "Claim inheritance" — the heir knows exactly what they're doing
- "Cancel vault & withdraw" — the owner can confidently exit

This is especially critical because DeadSwitch is non-custodial — there's no customer support to call if something goes wrong.

## Validation

Manifest validates against `specs/erc7730-v1.schema.json` (Draft 4 JSON Schema).

## Checklist

- [x] File placed at `registry/deadswitch/calldata-InheritanceVault.json`
- [x] Follows `calldata-` naming convention
- [x] Validates against ERC-7730 v1 schema
- [x] Single entity folder modified
- [x] Contract deployed with verifiable address
- [x] All functions have intents and labels
- [x] Complex types use proper formats (addressName, duration, raw)

---

Built for ETHGlobal Cannes 2026. Feedback welcome!


Lastoneparis:
Thanks for the automated feedback — all of it is addressed in 7c6d1f9, along with the missing test file.

**Recommendations**
- `interpolatedIntent` added to all 9 formats. Transaction values are embedded where they help the signer (`New heir {#._newBeneficiary}`, `Set guardian {#._guardian}`, `New interval {#._newInterval}`, `Extend {#._additionalSeconds}`). For `claim` I left the `bytes32` nullifier out of the sentence on purpose — it renders as a 66-character hex string, so it stays a labelled field instead.
- Deprecated `context.contract.abi` removed. The format keys already carry the full signatures.
- Intents kept to 30 characters and labels to 20, so nothing is truncated on a Ledger screen.

**Also in this commit**
- Migrated the descriptor from `erc7730-v1.schema.json` to v2 — it was the last v1 descriptor in the registry, and `interpolatedIntent` only exists in v2. As a consequence `required`/`excluded` became `visible: "never"` on the heir's World ID nullifier, and `metadata.info.legalName` was dropped (v2 `info` allows only `url` and `deploymentDate`).
- Added `registry/deadswitch/testsv2/calldata-InheritanceVault.tests.json`: 9 cases, one unsigned Sepolia transaction per function, with a `dataProvider` supplying local names for the heir and guardian addresses so name resolution is exercised.

I ran both reference runners locally at the commits pinned in the workflow before pushing — `sourcifyeth/clear-signing-test-runner` and `llbartekll/cs-test`, 9/9 on each — and CI now agrees.

One open point for your judgement: the vault at `0xF957cDA1f676B9EAE65Ab99982CAa3a31A193CB7` is not verified on Sourcify, so `erc7730 lint` emits a "could not fetch ABI" warning and cannot cross-check the display fields against a real ABI. Happy to get the contract verified on Sourcify if you would rather have that before merging.

kuzdogan:
Thanks for the submission. CI is green, but a manual check of the deployed contract turned up problems that CI cannot see.

**1. The contract is unverified.** `0xF957cDA1f676B9EAE65Ab99982CAa3a31A193CB7` on Sepolia has no source on Sourcify or Etherscan, so the ABI could only be checked against bytecode.

**2. Four of the nine functions in the descriptor do not exist in the deployed bytecode.** Comparing the descriptor's `display.formats` selectors with the 17 selectors present in the contract code:

| Function | Selector | In bytecode |
|---|---|---|
| `heartbeat()` | `0x3defb962` | ✅ |
| `claim(bytes32)` | `0xbd66528a` | ✅ |
| `cancel()` | `0xea8a1af0` | ✅ |
| `updateBeneficiary(address,bytes32,string)` | `0x26a06c39` | ✅ |
| `updateInterval(uint256)` | `0xd0f07926` | ✅ |
| `setGuardian(address)` | `0x8a0dac4a` | ❌ |
| `pauseRecovery()` | `0x7ef73896` | ❌ |
| `resumeFromRecovery()` | `0x3d3b3388` | ❌ |
| `extendHeartbeat(uint256)` | `0x7b29dcf0` | ❌ |

It looks like the descriptor was written against a newer version of the contract than the one at this address.

**3. The test fixtures are not real transactions.** All nine `rawTx` values are unsigned EIP-1559 envelopes (no signature fields), there are no `txHash` entries, and the contract has only one transaction on-chain — its deployment on 2026-04-03. The runners decode calldata only, which is why the four cases for non-existent functions still pass.

**To move this forward:**
- Verify the contract source on Sourcify (https://sourcify.dev) so the ABI can be checked.
- Either remove the four functions that are not deployed, or redeploy the version that has them and update the address.
- Replace the fixtures with real signed transactions sent to the contract, including their `txHash`.

Since this is a Sepolia-only deployment with no usage yet, it may also be worth waiting until there is a mainnet (or actively used) deployment before adding it to the registry.

Lastoneparis:
Thanks for the careful review — you're right on all three points, and I verified each one independently before replying.

**Root cause of the missing functions.** The vault at `0xF957cDA1f676B9EAE65Ab99982CAa3a31A193CB7` was deployed on 2026-04-03. The four functions you flagged (`setGuardian`, `pauseRecovery`, `resumeFromRecovery`, `extendHeartbeat`) were added the next day, 2026-04-04, in a commit that also added `submitActivityProof`. So the descriptor was written against a source tree one commit ahead of what is actually on-chain. I confirmed this locally: the Hardhat build artifact from the deployment run is a byte-for-byte match with the deployed runtime bytecode, metadata hash included.

**The fixtures.** Confirmed — all nine `rawTx` values RLP-decode to 9-field EIP-1559 payloads rather than the 12 fields a signed transaction carries, and none has a `txHash`. They were hand-assembled calldata envelopes, not real transactions. That was the wrong way to build a test file and it is exactly why the four non-existent functions still passed CI. My mistake, and thank you for catching it.

**Why I'm closing rather than fixing.** Digging into your last point surfaced something more fundamental. In this product each user deploys their own vault — the frontend calls `useDeployContract` with the vault bytecode directly, there is no factory and no shared instance. So every real user vault lives at a different address, and the one address in this PR is a single dormant demo vault (deployed once, never used; `getTimeRemaining()` has been 0 for weeks with no upkeep ever called). Pinning one Sepolia address in the registry does not describe what users would actually be signing against, and I do not have a stable address that would.

Closing this until there is a deployment worth registering — realistically a factory or shared implementation on mainnet with actual usage. If that happens I'll open a fresh PR with a verified contract and fixtures taken from real signed transactions. Appreciate you taking the time to check the bytecode by hand instead of trusting the green CI.

PR #2542, opened 2026-04-29
files moved in from llbartekll/clear-signing-erc7730-registry 

elasticroentgen:
as per request from hesterb via telegram


llbartekll (approved):

AmazedCurieMKE (approved):
approving to unblock the PR

hesterbruikman:
@elasticroentgen  can you tell what causes the check to fail? It seems to want to enforce labels

elasticroentgen:
hmm seems it wants one of those lables:
- documentation
- specifications
- descriptors
- ci
- tools

what would be the appropriate one?

hesterbruikman:
> hmm seems it wants one of those lables:
> 
> * documentation
> * specifications
> * descriptors
> * ci
> * tools
> 
> what would be the appropriate one?

I'd go with `tools`, could go for descriptors or documentation as well, but it's not essential for descriptors or documentation. To my understanding it's a helper file for the libraries to make sense of the registry.

PR #2543, opened 2026-04-30
Add auditors folder

hesterbruikman:
- Add folder for auditors. Auditors can create a PR to add themselves as known auditor. This aids discovery of descriptorHash by auditor. Auditors and Wallets can also agree on other discovery mechanisms such as ENS.
- Add README for auditors. Describes review, attestation and revocation process for auditors.

Note that this file might require updates in the coming weeks based on feedback and finalization of [ERC-8176 Descriptor Attestations for ERC-7730](https://github.com/llbartekll/ERCs/blob/bcb203b41cf2891d7959df160069a997cb567913/ERCS/erc-8176.md).


auditors/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+---
+
+## Key rules
+
+- Sign only after completing the full review
+- Sign the **exact file version** you reviewed
+- Never modify an existing attestation — issue a new one
+- If issues are found: **do not sign** — open a GitHub issue instead
+- If your key is compromised or you retract an attestation: submit a revocation on-chain via EAS
```
manuelwedler, line 118: In case of a key-compromise, wallets should be notified, such that the key is removed from their list of trusted keys.

Ofc revocation could still be done additionally, but might be cumbersome when a lot of attestations have been made.

auditors/README.md
```diff
@@ -0,0 +1,91 @@
...
+}
+```
+
+`descriptorHash = keccak256(RFC 8785 JCS-canonicalized descriptor JSON)` — do not hash raw file bytes.
+
+**Publish:** Download the attestation `.json` from EAS and submit it via PR to the registry at:
+
+```
+registry/<project>/sigs/<descriptor-name>/eip155-1-0xYourAddress.json
+```
```
manuelwedler, line 107: We need to update this according to the updated ERC-8176.

I think auditors only need to calculate the `descriptorHash` and can then sign via the EAS frontend. It would be helpful to add the link to the EAS schema here (https://easscan.org/schema/view/0xe023eef113c1670774801c34b377fdf612dd8a4d2fa92fe382e15bd91fafb5c2). Auditors can click "Attest with schema" and fill the form. They should select offchain in the form.

@llbartekll any more input from your side about the signing process?

auditors/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+**Your role:** Review ERC-7730 descriptors — the JSON files that tell Ethereum wallets what to display when users sign transactions — and publish a cryptographic attestation confirming your review.
+
+---
+
+## What you're committing to
+
+- Review assigned descriptors within **7 days** of being tagged
+- Publish a signed attestation (or open an issue) for each descriptor reviewed
+- Maintain your attestation as descriptors evolve — new version = new attestation required
```
manuelwedler, line 11: Some aspects were not entirely clear to me after fully reading:

- Do we assume PR submitters merge by themselves now? I assume yes, since we have aligned on EAS now.
- Where do auditors get tagged? Do we use the list of auditors and tag all auditors on every new PR (via  CI)?

llbartekll, line 11: I don't think we've discussed descriptor versioning yet. It's not a problem now but if descriptors can be edited in place, all existing attestations become stale. the descriptor hash changes, so the attestations no longer match.

auditors/README.md
llbartekll: The process you described Manual is exactly right!

auditors/README.md
manuelwedler: The issue I see here is that the libraries won't be able to fetch the previously attested version from GitHub anymore, when the descriptor is changed on GitHub. On top of that, if anyone can merge, a malicious actor can simply break clear signing for a contract by updating a descriptor (the descriptor wouldn't have attestations, so transactions won't be formatted).

llbartekll: > a malicious actor can simply break clear signing for a contract by updating a descriptor

I think anyone can merge attestation but not descriptor change. Descriptor change needs to be reviewed and submitted by the contract "owner"

> The issue I see here is that the libraries won't be able to fetch the previously attested version from GitHub

unless we store descriptors versioned and attestations per version -- tho it may become not trivial to maintain, definitely an item for the next WG call

hesterbruikman: - Do we assume PR submitters merge by themselves now? I assume yes, since we have aligned on EAS now.
This is the proposal. We will need to set up logic to mitigate poisoning like using an `automerge` label that needs to be applied by a maintainer. This should not affect the auditor guidance though.

- Where do auditors get tagged? Do we use the list of auditors and tag all auditors on every new PR (via CI)?
Great point, tagging doesn't make sense in the new attestation model. Updated to 
"Review descriptors regularly, e.g. by adding this repo to your [Watch list](https://github.com/watching) and setting up [Notifications](https://github.com/settings/notifications)"

auditors/README.md
hesterbruikman: Added:
"- If you have explicitly shared your key with wallets, notify them of the compromise"

auditors/README.md
hesterbruikman: Noted for the agenda!

auditors/README.md
hesterbruikman: updated to 

```
After a full review passes, create an EAS offchain attestation (ERC-8176 schema):

{
  "$type": "8176/v1",
  "descriptorHash": "0x...",
  "attester": "eip155:1:0xYourAddress",
  "attesterType": "EOA",
  "issuedAt": <unix timestamp>,
  "expiresAt": null,
  "signature": "0x..."
}
Calculate the descriptor hash
descriptorHash = keccak256(RFC 8785 JCS-canonicalized descriptor JSON) — do not hash raw file bytes.

Go to [Ethereum Attestation Service](https://easscan.org/schema/view/0xe023eef113c1670774801c34b377fdf612dd8a4d2fa92fe382e15bd91fafb5c2), select 'Attest with schema' and use 'Offchain'
```

Is there any more instruction needed on the calculate descriptorHash?

PatrickAlphaC:
Do we/should we add anything here about upgrades/changes? Let's say an auditor files a report here, and then the contract upgrades later. Now that report is outdated, but the auditor is still on the hook?

Looks fine otherwise, but I suspect we will probably tweak the process as we continue.

auditors/README.md
```diff
@@ -0,0 +1,88 @@
...
+{
+  "$type": "8176/v1",
+  "descriptorHash": "0x...",
+  "attester": "eip155:1:0xYourAddress",
+  "attesterType": "EOA",
+  "issuedAt": <unix timestamp>,
+  "expiresAt": null,
+  "signature": "0x..."
+}
+```
```
llbartekll: this envelope doesn't exist anymore

auditors/README.md
hesterbruikman: concluded on call that versioning affects descriptor file name, which is adopted in, but does not affect the auditor workflow.

auditors/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+## The 4-step review
+
+**1. Project check**
+Find the project URL in the descriptor JSON. Confirm the protocol's purpose and that the PR submitter is plausibly affiliated.
+
+**2. Contract verification**
+Confirm the contract is verified at [repo.sourcify.dev](https://repo.sourcify.dev). Cross-check the contract address and ABI against the descriptor. If unverified — **do not sign**.
+
+**3. Descriptor accuracy**
```
PatrickAlphaC, line 46: Upgradeability? Should we be commenting on that somewhere? Or leave that off. Feel like a lot of extra work tbh... What do others think?

auditors/README.md
```diff
@@ -0,0 +1,83 @@
...
+- Parameter names, types, and ordering match the ABI
+- Function selectors are correct
+- Intent fields reflect real user impact — no misleading simplifications
+- Approvals, transfers, and privileged actions are correctly flagged
+- Payable vs non-payable is accurate
+- Cross-chain deployments are consistent
+
+**4. Tester validation**
+Run the descriptor through the Tester tool to simulate transactions and confirm the output is unambiguous.
```
PatrickAlphaC: What is "the tester tool"?

auditors/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+After a full review passes, create an **EAS offchain attestation** (ERC-8176 schema):
+
+- Calculate the descriptor hash
+
+`descriptorHash = keccak256(RFC 8785 JCS-canonicalized descriptor JSON)` — do not hash raw file bytes.
+
+- Go to [Ethereum Attestation Service](https://easscan.org/schema/view/0xe023eef113c1670774801c34b377fdf612dd8a4d2fa92fe382e15bd91fafb5c2), select 'Attest with schema' and use 'Offchain'
+- Sign the attestation
+- Copy and save the raw attestation data on EAS as json file and submit it via PR to the registry at:
```
PatrickAlphaC, line 103: Can we get an example of this?

PatrickAlphaC (commented):
Left comments on the notes. I think it would be helpful to do a walkthrough of the process once as well.

auditors/README.md
llbartekll: https://easscan.org/offchain/attestation/view/0x63ed601e0f66f3dee39015a0a0450de45c2b064f600e48752ac22109598eb8b2

auditors/README.md
```diff
@@ -0,0 +1,83 @@
...
+- Review descriptors regularly, e.g. by adding this repo to your [Watch list](https://github.com/watching) and setting up [Notifications](https://github.com/settings/notifications)
+- Publish a signed attestation (or open an issue) for each descriptor reviewed
+- Maintain your attestation as descriptors evolve — new version = new attestation required
+
+---
+
+## The 4-step review
+
+**1. Project check**
+Find the project URL in the descriptor JSON. Confirm the protocol's purpose and that the PR submitter is plausibly affiliated.
```
PatrickAlphaC, line 41: Ideally, we have a more objective way to check this. It's a bit hand-wavey at the moment.

auditors/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+**2. Contract verification**
+Confirm the contract is verified at [repo.sourcify.dev](https://repo.sourcify.dev). Cross-check the contract address and ABI against the descriptor. If unverified — **do not sign**.
```
PatrickAlphaC: Historically, contract verification was not required; moving forward, it is a requirement. If the chain is not EVM equivalent or supported by Sourcify, there needs to be a comment/message in the PR showing where the contract is verified.

auditors/README.md
PatrickAlphaC, line 46: It might be good to have the protocols also put a link to documentation in the spec. There is a metadata field where they can do that. Like a `docs` feild in the `info` feild.

PatrickAlphaC, line 46: Auditors can choose their level of thoroughness. I think just looking at intent to make it "technically" accurate is good enough. It's too much to ask to have an auditor do a security review to make sure the intent matches in all cases.

PatrickAlphaC: We should add some docs around upgrades... Different wallets might have their own implementations for deteching intent upgrades.

PatrickAlphaC: This is a real threat, we will need some infra around it.

PatrickAlphaC: On second thought, auditors will need to do full reviews to check for the ability for the protocol to rug users by changing intents.

auditors/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+---
+
+## Submitting your attestation
+
+After a full review passes, create an **EAS offchain attestation** (ERC-8176 schema):
+
+- Calculate the descriptor hash
+
+`descriptorHash = keccak256(RFC 8785 JCS-canonicalized descriptor JSON)` — do not hash raw file bytes.
```
PatrickAlphaC, line 92: This will be the whole JSON file. We should have a tool to auto-generate/help.

auditors/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+- Go to [Ethereum Attestation Service](https://easscan.org/schema/view/0xe023eef113c1670774801c34b377fdf612dd8a4d2fa92fe382e15bd91fafb5c2), select 'Attest with schema' and use 'Offchain'
```
PatrickAlphaC, line 101: Put the descriptorHash in there

PatrickAlphaC (commented):
Left more notes. We have not quite solved the upgrades issue. Right now it's sort of just "the auditor needs to always be checking" which seems like... A lot. It's up to the wallets to choose how much warning they want to put for intent changes, ie "do you always warn that USDC could rug pull at any time?" etc.

PatrickAlphaC:
We should have a place where auditors can say what functions could have their intents change. The auditor should ask the questions:
1. Does the intent match what the function does?
2. Can the intent change in the future? If so, how?
And we have a section where auditors can have a list of possible intent changes / future risks.

hesterbruikman:
> Do we/should we add anything here about upgrades/changes? Let's say an auditor files a report here, and then the contract upgrades later. Now that report is outdated, but the auditor is still on the hook?
> 
> Looks fine otherwise, but I suspect we will probably tweak the process as we continue.

addressed by this [commit](efce125fae3d5ee65d6c52e24ab25410104c8ad5) from @PatrickAlphaC

auditors/README.md
hesterbruikman: Ultimately something like adding a signature with an org level ENS to the PR could work.

auditors/README.md
hesterbruikman: resolving as added in via your commit @PatrickAlphaC

auditors/README.md
hesterbruikman: forthcoming, resolving as discussed on TG

PatrickAlphaC:
Sorry for all my comments, this LGTM. There are a few TODOs, but they are outlined in the instructions, and this, at least, is a great baseline with a specific scope on an auditor's job.

AmazedCurieMKE (approved):

PR #2550, opened 2026-05-09
Add testnet deployment for Serenita's ETH Staking Vault

eth2353:
Adds the Hoodi testnet deployment of Serenita's EthVault staking smart contract.

I also unified the letter case across all intents.


manuelwedler:
Hey there :)

kuzdogan:
Manual check of the new Hoodi deployment `0xf92eff17ac8cef8e45dec00abdcd8788ecf83309` (chain 560048):

- It is an OpenZeppelin `ERC1967Proxy` whose implementation slot points to `0x927a83c679a5e1a6435d6bfaef7f20d4db23e2cc` — the same address as the mainnet `EthVault` implementation already covered by this descriptor.
- Both the proxy and the implementation are now fully verified on Sourcify (exact match on runtime and creation bytecode; implementation `contracts/vaults/ethereum/EthVault.sol:EthVault`, solc 0.8.26). Etherscan reports the same proxy/implementation pair.
- The implementation ABI contains every function this descriptor covers (`claimExitedAssets`, `deposit`, `enterExitQueue`, `multicall`, `updateState`, `updateStateAndDeposit`), and `version()` returns 5 on both chains.

So the address itself is correct and CI is green after the `testsv2` intent fix.

**Marking as blocked for now**, for a reason that is not specific to this PR: the descriptor is bound to the proxy address, while the functions it describes live in the implementation. An upgrade behind the proxy could change the ABI without any change to the address the descriptor matches on. We have not yet decided how the registry should handle proxy deployments in general (pin the implementation, record it alongside the proxy, re-verify on upgrade, or accept the risk). Once that policy exists, this PR can move forward without further changes.

eth2353:
> Marking as blocked for now, for a reason that is not specific to this PR: the descriptor is bound to the proxy address, while the functions it describes live in the implementation. An upgrade behind the proxy could change the ABI without any change to the address the descriptor matches on. We have not yet decided how the registry should handle proxy deployments in general (pin the implementation, record it alongside the proxy, re-verify on upgrade, or accept the risk). Once that policy exists, this PR can move forward without further changes.

Hi Kaan,

Thanks for the review and fixing the v2 tests.

I wanted to let you know that in this case we are unable to **arbitrarily** upgrade the contract to any implementation. The upgrade functionality is guarded by the StakeWise VaultsRegistry which has a DAO governance-controlled list of valid implementations - we can only upgrade the contract to one of those - see [here](https://github.com/stakewise/v3-core/blob/fc70cbe1b3d41bc5f78434830d837aa270ca33bc/contracts/vaults/modules/VaultVersion.sol#L39). This somewhat changes the trust assumptions, from Serenita onto the StakeWise DAO.

Ultimately, I don't think it materially changes much. I just wanted to make a note of it in case you want to account for this differently somehow.

In terms of a decision on the broader topic, I would personally be in favor of the most trustless variant possible while acknowledging the best option may not be achievable in practice due to time constraints. In our specific case, we are looking at roughly 1 implementation upgrade per year, in which case more strict variants may still be practical.

kuzdogan:
Hey there :) 

The AI response aside, this is something we need to resolve on our side. The spec already mandates to bind descriptors to implementations and not to the proxy. However, I acknowledge there are existing proxies in the registry... I'd kindly ask 1-2 weeks for us to reach a consensus on this. Maybe we'll just approve proxies for now and handle all of them with a deadline (ie. no proxies allowed after this date). It's still reassuring to know that arbitrary upgrades are not possible. But we also need reassurances on the spec side to catch any upgrades and consequently any descriptor changes.

kuzdogan:
@eth2353 So my understanding of the spec was half right. 

The spec mandates to bind to implementations for proxies with reference implementations such as SafeProxy's pointing to singleton references.

For upgradable proxies it's ok to bind the descriptor to the proxy. However this is still with the caveat that the implementation might change. For that we have https://github.com/ethereum/ERCs/pull/1738 on the way 

So this is good to go, thanks!

kuzdogan (approved):

PR #2553, opened 2026-05-11
fix: make NttManager relayingFee field optional

antoineborg:
Make the `NttManager.transfer` `relayingFee` field optional so the descriptor stays valid when `msg.value` is 0 (currently triggers `INVALID_DESCRIPTOR` and falls back to raw calldata on Ledger).
Test expectations updated to drop the now-hidden "Relaying Fee / 0 ETH" lines.


hesterbruikman (approved):
Thanks for updating!

PR #2556, opened 2026-05-12
Add Ekubo clear-signing descriptors and ve33 support

moodysalem:
## Summary

- Add ERC-7730 calldata descriptors for Ekubo's Router, Orders, and Positions contracts.
- Add user-facing ve33 coverage for VeToken, Ve33Positions, and Ve33Periphery on Ekubo and Ekubo Testnet (`4663` and `46630`).
- Cover the Router's single- and multi-multihop swap entrypoints and add the Ekubo deployment (`4663`) to the existing Router, Orders, and Positions descriptors.
- Add `interpolatedIntent` to every described function.
- Replace the legacy fixtures with one v2 fixture per descriptor.

The six descriptors cover 128 state-changing ABI entrypoints. The reserved `locked_6416899205(uint256)` callback is intentionally omitted, as are read-only functions such as `quote`.

## Scope decisions

The core Ve33 extension is not a normal user-facing call target: its external surface is primarily Core hooks and permissionless accounting maintenance. User actions are represented by VeToken, Ve33Positions, and Ve33Periphery instead.

I also evaluated Ekubo's Yul router. Its calldata is a selectorless, packed command format whose first four bytes are route data rather than an ABI function selector. The current ERC-7730 calldata format selects displays by ABI signature/selector, so describing it as a conventional contract call could render the wrong intent. It is intentionally left out until the registry format can safely identify this encoding.

## Validation

- Rebased on current `master` (`8619a90`).
- Canonically formatted with `erc7730 format`.
- All descriptors and v2 fixtures pass their repository JSON schemas.
- `erc7730 lint` passes; remaining warnings are limited to the reserved lock callback, the read-only Router quote, and Etherscan not supporting chain IDs `4663`/`46630`.
- All six v2 fixtures pass both the pinned Sourcify TypeScript runner and the pinned Rust clear-signing runner.
- Every descriptor signature was checked against locally compiled contract ABIs with no mismatches.
- `forge snapshot --offline`: 964 tests passed, 0 failed.


hesterbruikman:
@moodysalem Thanks for adding these descriptors! I noticed you're not using interpolatedIntent. Is that intentional/not relevant for your protocol?

It's not required, just wondering.

"The interpolatedIntent key is a suggested short string representation of intent and fields that wallet CAN use instead of having a dedicated layout."

manuelwedler:
Hey there :)

moodysalem:
@hesterbruikman @manuelwedler Thanks for the feedback — everything requested is now addressed:

- Added `interpolatedIntent` to all 128 described functions, including assertions in the v2 fixtures.
- Rebased the PR onto the latest `master` (`8619a90`).
- Removed all three legacy `tests/` fixtures and added `testsv2/` fixtures for every changed descriptor (six total).
- Ran all six fixtures successfully against both the Sourcify TypeScript and Rust clear-signing implementations.

While updating this, I also added the newly deployed user-facing ve33 contracts (VeToken, Ve33Positions, and Ve33Periphery), Ekubo-chain deployments for the existing contracts, and the Router's missing multihop entrypoints. The descriptors match the locally compiled ABIs, schemas/lint pass, and the EVM suite passes all 964 offline tests.

I investigated the Yul router too, but did not add an unsafe descriptor: it uses selectorless packed calldata, so its first four bytes are route data rather than an ABI selector and the current ERC-7730 lookup model cannot identify it reliably. The PR description now documents that limitation and the full validation performed.

This should be ready for another review. Because this is a fork PR, the clear-signing workflow may need the existing `run-tests` label retriggered after this force-push.

moodysalem:
@hesterbruikman I'm building and testing something that makes use of clear signing, could you give this another look soon?

manuelwedler:
reopened to trigger CI. But also worth rebasing on master due to latest CI changes

moodysalem:
@manuelwedler Rebased onto latest `master` (now on `bd8dd9a`), dropping the earlier merge commit so the branch is linear with five commits on top of master. The `registry/ekubo` tree is byte-identical to what CI last validated green (all Sourcify + Rust cases passing), so this is purely a history/base update. Nothing else outstanding on my side — `interpolatedIntent` and the v2 test fixtures from the earlier round are all in. Ready for review whenever you have a moment.

moodysalem:
@manuelwedler @hesterbruikman Friendly ping on this one — it's been sitting green and ready since the last round of feedback was addressed.

Current state:
- Rebased again onto latest `master` just now, so it picks up the recent CI changes (including the pinned Python tooling) and is 0 commits behind.
- All checks passing on the new head: both the Sourcify TypeScript and Rust suites, plus schema/index/filename/attestation validation.
- Everything from the earlier review is in: `interpolatedIntent` on all described functions, legacy `tests/` removed, `testsv2/` fixtures added for all six descriptors.
- No open review threads on my side.

Anything else you'd like changed, or is this good to merge?

fbwoolf (approved):
Approving. Two notes, neither blocking.

**Chain 80002 has no contract at the router address.** While verifying the deployments on Sourcify I
submitted the 21 unverified entries to `/v2/verify/similarity`; 20 came back verified, and
`80002:0xd26f20001a72a18C002b00e6710000d68700ce00` was rejected with `cannot_fetch_bytecode`.
`eth_getCode` returns `0x` there on publicnode, dRPC and Tenderly. Since the router is deployed
deterministically from `0x00000C771F6176268D5A9846E0956C3eF58597A1` on every other chain, the address
cannot hold different code later, so this is an accuracy fix rather than a risk: drop the Polygon
Amoy entry, or deploy there.

Worth re-triggering CI either way. On the current run the linter could not fetch a reference ABI for
chains 1, 4326 and 4663, so `Orders`, `Positions` and the Ve33 descriptors had their display fields
validated against no ABI at all. Those addresses are verified now. For what it is worth I checked all
128 format keys against the deployed ABIs by hand: selector, full type signature and parameter names
match with no exceptions, and every deployment of a given descriptor exposes an identical signature
set.

**14 `interpolatedIntent` values do not reflect their intent.**

| pattern | count | where |
| --- | --- | --- |
| `Action for {id}` | 11 | `Orders` and `Positions`: `burn`, `approve`, `transferFrom`, `safeTransferFrom` (x2), `executeVirtualOrdersAndGetCurrentOrderInfo` |
| `Send to {recipient}` | 3 | two router swaps, `Positions.withdrawProtocolFees` |

A position-NFT `transferFrom` summarises as "Action for 1234", which drops both the verb and the
recipient, and a swap summarising as "Send to 0x..." names the wrong action. `Ve33Positions` already
handles the same NFT surface correctly ("Transfer NFT {id}", "Approve NFT {id}",
"Burn Ve33 position NFT {id}"), so copying that pattern into the two older descriptors covers it.

moodysalem:
@fbwoolf Thanks for the careful review — both notes are addressed in `2dd4050`.

**1. Chain 80002 — dropped.** Confirmed independently before removing: `eth_getCode` at `0xd26f...ce00` returns `0x` on publicnode, dRPC and Tenderly, and the `evm-contracts` broadcast record has no router deployment for Amoy (its `DeployAll` run deployed Core/TWAMM/Orders/Positions there, but never the router). Removed rather than deployed, so the descriptor matches what is actually on chain today.

**2. All remaining addresses re-verified against the contracts repo.** Since you had flagged one bad entry, I checked every `(chainId, address)` pair in the six descriptors rather than just that one — 43 pairs, cross-referenced against the `evm-contracts` deploy scripts and broadcast receipts and then confirmed with `eth_getCode`:

| result | count |
| --- | --- |
| bytecode present | 42 |
| empty (`0x`) | 1 — the Amoy router, now removed |

Code size is identical for every deployment of a given contract (router 11,122 B on all 18 remaining chains; Positions 21,978 B; Orders 13,528 B; VeToken 18,491 B; Ve33Positions 21,585 B; Ve33Periphery 2,569 B), which matches the deterministic-CREATE2 story — `script/DeployRouter.s.sol` pins salt `0x38e2e731…3982`.

Two things worth recording, since neither is obvious from the descriptors:
- **4326 is MegaETH, not a Robinhood chain.** Robinhood is 4663 (mainnet) and 46630 (testnet) — I confirmed via `eth_chainId` against the RPCs in Ekubo's own interface config. The broadcast directory for 4326 does not record the Orders/Positions deployments, but both addresses hold the expected code on chain, so the entries are correct and the broadcast record is simply incomplete there.
- **42431 is Tempo Testnet Moderato**, and its router is live with the same 11,122 B — keeping it. Its broadcast run has no receipts, so I verified this one on chain rather than trusting the repo.

**3. The 14 `interpolatedIntent` values are rewritten**, following the `Ve33Positions` pattern you pointed at:

| function | was | now |
| --- | --- | --- |
| `Orders.burn` / `Positions.burn` | `Action for {id}` | `Burn order NFT {id}` / `Burn position NFT {id}` |
| `approve` (both) | `Action for {id}` | `Approve order NFT {id}` / `Approve position NFT {id}` |
| `transferFrom`, `safeTransferFrom` x2 (both) | `Action for {id}` | `Transfer order NFT {id}` / `Transfer position NFT {id}` |
| `Orders.executeVirtualOrdersAndGetCurrentOrderInfo` | `Action for {id}` | `Refresh DCA order {id}` |
| `Router.swap` | `Send to {recipient}` | `Swap, send to {recipient}` |
| `Router.swapAllowPartialFill` | `Send to {recipient}` | `Partial swap to {recipient}` |
| `Positions.withdrawProtocolFees` | `Send to {recipient}` | `Withdraw fees to {recipient}` |

One deliberate choice: my first pass named the counterparty too (`Transfer position NFT {id} to {to}`, `Approve order NFT {id} for {account}`), but `erc7730 lint` warns those exceed the 30-character Ledger budget, so the NFT summaries carry the verb and the token id and leave the recipient to the fields — which I think is also why `Ve33Positions` reads the way it does. Happy to add the recipient back if you would rather have it and accept the truncation.

Verified locally before pushing: `erc7730 lint` produces an identical warning set to before the change (the four pre-existing "missing display format" notes for `locked_6416899205` and `quote`, nothing new), all 12 files validate against `erc7730-v2` / `erc7730-tests-v2`, and `generate-index.js --validate` passes.

Not addressed, as follow-up rather than scope creep: there is a second Orders/Positions implementation family (`0x9bb520b6…` / `0xa2971e0c…`) deployed on ~22 chains with no descriptor. Happy to open a separate PR for those.

PR #2561, opened 2026-05-13
Add TeraSwap FeeCollector V2 clear signing descriptor

TeraHashAlpha:
# Add TeraSwap FeeCollector V2 descriptor

This PR adds an ERC-7730 descriptor for **TeraSwap FeeCollector V2**, the
on-chain entry point for every swap routed through TeraSwap's
meta-aggregator (Ethereum mainnet).

## Contract

| Field | Value |
|---|---|
| Name | TeraSwap FeeCollector V2 |
| Chain | Ethereum mainnet (chainId 1) |
| Address | `0x47f24068932Ac49bcbeD3aD105af57C6ECDF7459` |
| Etherscan | https://etherscan.io/address/0x47f24068932Ac49bcbeD3aD105af57C6ECDF7459#code |
| Source | https://github.com/TeraHashAlpha/teraswap |

## Functions covered

| Function | Selector | Purpose |
|---|---|---|
| `swapTokenWithFee(address,uint256,address,bytes,address,uint256)` | `0x7f7663d4` | ERC-20 input swap |
| `swapETHWithFee(address,bytes,address,uint256)` | `0x7739563c` | Native-ETH input swap |

Admin functions are deliberately out of scope — only the two
user-signing functions are described. `routerData` (encoded inner-DEX
calldata) is hidden via `visible: "never"` on both formats.

## Verification

- Function selectors verified against deployed bytecode (Etherscan verified source)
- Descriptor validates with `erc7730 lint` — 0 errors
- CI test in source repo re-computes selectors on every build and cross-references with frontend ABI
- JSON validates against `specs/erc7730-v2.schema.json`


TeraHashAlpha:
Friendly bump on this one 🙏  
validate descriptors and validate JSON schemas are both green. The ERC-7730 Analysis check was cancelled on a 1440-minute timeout on the runner side rather than failing on the descriptor itself. Is there anything needed from my end (a re-run, a rebase, or a doc change) to get this into the review queue? Happy to make any adjustments. Thanks!

manuelwedler:
Hey there :)

TeraHashAlpha:
Rebased on latest master (real `git rebase`, no conflicts) and converted the FeeCollector V2 tests to the new `testsv2/` v2 format with two real mainnet transactions (`swapTokenWithFee`, `swapETHWithFee`), removing the old placeholder-format test file — validated with `erc7730 lint` (descriptor) and `check-jsonschema` against `erc7730-tests-v2.schema.json` (test file), both green. Ready for another look, thank you!

TeraHashAlpha:
Rebased onto upstream/master and converted the FeeCollector V2 tests to the v2 format (testsv2/, old-format tests removed) — ready for re-review. Note: local erc7730 lint currently errors on our tests file, but the same error reproduces on the already-merged midas testsv2 file, so it looks like a stale-CLI issue rather than content; the file validates cleanly against specs/erc7730-tests-v2.schema.json.

fbwoolf (changes requested):
Two things to fix, one of them blocking.

**`@.value` uses the wrong formatter, which is why both runners fail the `swapETHWithFee` case.** `tokenAmount` derives its magnitude and ticker from the `token` or `tokenPath` parameter. This field sets neither, so `nativeCurrencyAddress` has nothing to match against and both runners fall back to the raw integer `50000000000000000`. The native-currency formatter is `amount`, which every other `@.value` field in the registry uses:

```json
{ "path": "@.value", "label": "ETH Amount", "format": "amount" }
```

The existing `"0.05 ETH"` expectation should pass unchanged after that.

**`tokenOut` of `address(0)` is not handled.** The contract treats it as native ETH output, and selling a token for ETH through `swapTokenWithFee` is an ordinary path. Today that renders "Output Token" as the zero address and resolves "Minimum Output" against a token at `0x0`. Adding `"nativeCurrencyAddress": "0x0000000000000000000000000000000000000000"` to the `minimumOutput` params covers the amount side. Not blocking.

On the earlier lint note: `erc7730 lint` only takes descriptors, and the README's validate-all command excludes `*/testsv2/*` and `*.tests.json`. The error on your tests file and on the midas one is the wrong input rather than a stale CLI, so there is nothing to chase there.

TeraHashAlpha:
Pushed a fix commit addressing both review comments: @.value on swapETHWithFee now uses the amount format instead of an unresolvable tokenAmount, and minimumOutput on swapTokenWithFee now carries nativeCurrencyAddress so tokenOut == address(0) resolves as native ETH. Ready for re-review.

fbwoolf (approved):

PR #2567, opened 2026-05-13
ci: nightly on-chain selector drift check

0xbeekeeper:
## Summary

`tools/scripts/check-contract-functions.js` already validates that a descriptor's function selectors exist in the on-chain ABI (with proxy resolution). Today it's only runnable manually, so descriptors can become stale when their backing contracts upgrade, without CI noticing.

This PR adds:

- **`tools/scripts/run-drift-check-all.sh`** — iterates every `calldata-*.json` descriptor under `registry/**` and `ercs/`, runs the checker per-file, and emits a markdown summary. Exits 0 on a clean run, 2 if any descriptor has mismatches or errors. Writes to `$GITHUB_STEP_SUMMARY` when present.

- **`.github/workflows/nightly-drift.yml`** — runs daily at 03:00 UTC and on `workflow_dispatch`. **Skips cleanly when `ETHERSCAN_API_KEY` is not set on the repo**, so this workflow is safe to merge even before the secret is configured. When drift is detected, opens (or updates) a sticky issue titled `🌙 Nightly drift report` with the summary, and uploads `drift-report.md` as a 30-day artifact.

## Why nightly + sticky issue

- Daily is the right cadence: contract upgrades are rare-but-real (proxy implementations rotate, parameters change) and same-day notice is high signal.
- A sticky issue beats a fresh issue per night: maintainers see one item to act on, not 30 stale duplicates, and old runs leave history via comments.
- The script is also re-runnable via `workflow_dispatch` for spot checks during incident response.

## To enable on the upstream repo

1. Merge this PR.
2. Add `ETHERSCAN_API_KEY` to the repo's **Actions secrets** (the README already references this convention for `tools/analyzer/` and `tools/scripts/`).

Until step 2, every nightly run is a no-op (one notice line in the summary, no other side effects).

## Test plan

- [x] `bash -n tools/scripts/run-drift-check-all.sh` — shell syntax check passes.
- [x] YAML parses cleanly.
- [ ] Trigger via `workflow_dispatch` from a fork with `ETHERSCAN_API_KEY` set and confirm the report renders + the sticky issue is created.
- [ ] Trigger from a repo without the secret and confirm the gate step prints the notice and remaining steps are skipped.

### DIFF
--- a/.github/workflows/nightly-drift.yml
+++ b/.github/workflows/nightly-drift.yml
@@ -1,0 +1,90 @@
+name: 🌙 nightly selector drift
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Every day at 03:00 UTC
+    - cron: "0 3 * * *"
+
+permissions:
+  contents: read
+  issues: write
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
+
+jobs:
+  drift:
+    name: 🔬 check on-chain selectors
+    runs-on: ubuntu-latest
+    timeout-minutes: 180
+    env:
+      ETHERSCAN_API_KEY: ${{ secrets.ETHERSCAN_API_KEY }}
+    steps:
+      - name: Skip when no Etherscan API key
+        if: ${{ env.ETHERSCAN_API_KEY == '' }}
+        run: |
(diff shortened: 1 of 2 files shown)


manuelwedler:
The linter CI jobs already do this.

PR #2568, opened 2026-05-13
ci: add PR-time on-chain selector check (warning-only)

0xbeekeeper:
> Stacked on #2567 (nightly drift workflow). Please review that one first; this branch is based on it. Once #2567 lands, the diff here will resolve to the single new job in `pull_request.yml`.

## Summary

Adds a new job `validate_selectors_warning` to `.github/workflows/pull_request.yml` that runs `tools/scripts/check-contract-functions.js` against any `calldata-*.json` descriptors modified in the PR.

The job is **intentionally warning-only**:

- `continue-on-error: true`
- the final `exit 0` guarantees the job never reports failure
- skips cleanly when `ETHERSCAN_API_KEY` is not set (typical for fork-originating PRs without access to secrets)
- the authoritative gate stays the nightly drift workflow (#2567)

When a selector mismatch is detected, the job emits:

- a `::warning file=path/to/descriptor.json::Selector mismatch with on-chain ABI` annotation on the PR diff
- a collapsible `<details>` block in the job summary with the full checker output

## Why warning-only

- Hard-gating selector checks at PR time would block legitimate work whenever Etherscan is flaky, when the secret is rotating, or when a fork PR can't access the secret.
- It also lets maintainers see one round of real-world output before deciding to flip it to a required check later (just remove `continue-on-error` and the trailing `exit 0`).
- The nightly job already catches drift at a stable cadence — this PR-time check is purely a UX nicety that pulls forward the signal during review.

## To enable on the upstream repo

Same as #2567: add `ETHERSCAN_API_KEY` to the repo's Actions secrets. Until then the job emits a notice and skips.

## Test plan

- [x] YAML parses cleanly (`python3 -c 'import yaml; yaml.safe_load(open(...))'`).
- [ ] Open a test PR that modifies a known-good calldata descriptor and confirm the job summary shows ✅.
- [ ] Open a test PR with a deliberately broken selector and confirm the warning annotation lands on the descriptor file.
- [ ] Open a fork PR with no secret access and confirm the gate step prints the notice and the rest is skipped.

### DIFF
--- a/.github/workflows/nightly-drift.yml
+++ b/.github/workflows/nightly-drift.yml
@@ -1,0 +1,90 @@
+name: 🌙 nightly selector drift
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Every day at 03:00 UTC
+    - cron: "0 3 * * *"
+
+permissions:
+  contents: read
+  issues: write
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
+
+jobs:
+  drift:
+    name: 🔬 check on-chain selectors
+    runs-on: ubuntu-latest
+    timeout-minutes: 180
+    env:
+      ETHERSCAN_API_KEY: ${{ secrets.ETHERSCAN_API_KEY }}
+    steps:
+      - name: Skip when no Etherscan API key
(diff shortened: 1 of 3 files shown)


manuelwedler:
The linter CI jobs already do this.

PR #2570, opened 2026-05-14
Add Flying Tulip clear signing descriptors

patcito:
Adds clear signing support for Flying Tulip wallet interactions across put, ftUSD, lend, and spot.

Includes direct calldata descriptors, session creation support, relevant EIP-712 signing flows, and Permit2 support on Sonic.

Session execution transactions are intentionally left out because those are relayer or delegate flows, not wallet prompts.


hesterbruikman:
Thanks for contributing to the registry! I noticed that the descriptors do not include interpolatedIntent. Is that intentional/not relevant for the contract?

Recommended, not required

"The interpolatedIntent key is a suggested short string representation of intent and fields that wallet CAN use instead of having a dedicated layout."

patcito:
Hey @hesterbruikman, added  interpolatedIntent to the Flying Tulip descriptors and fixed the schema validation issue from the nested array fields. Also added the Flying Tulip test fixtures in the previous update.

Checks should be green now.

hesterbruikman (approved):
Thanks for adding to the registry!

PR #2586, opened 2026-05-21
Common Test Strategy

manuelwedler:
## Summary

- **Goal:** Introduces a multi-runner reference-test pipeline so any conformant clear-signing implementation can be validated against the registry, alongside Ledger's existing tests.
- **Schema:** New v2 test schema (`specs/erc7730-tests-v2.schema.json`) replaces the legacy flat `expectedTexts` array with a structured `{ intent, interpolatedIntent?, owner, fields }` shape that mirrors what wallets render. Tests live in `registry/<entity>/testsv2/` and reference their descriptor explicitly. An optional `dataProvider` block supplies mock token metadata and address-name lookups so runners don't need network access.
- **Runner output:** Unified runner output spec under `.github/test-results/` defines the single `results.json` shape every runner emits (`runner`, `implementation`, `cases[].{status, rendered|error}`), with `pass`/`fail`/`error`/`skipped` examples.
- **CI:** Three composite actions (`run-ledger-tests`, `run-sourcify-tests`, `run-rust-tests`) run in parallel on every PR. Results are collated into a single PR comment with one column per implementation, showing per-case pass/fail and a diff against expected output. The workflow fails when any case fails or when a touched descriptor has no `testsv2/` file.
- **Gating:** Cross-repo PRs require a maintainer to add a `run-tests` label before any runner executes; same-repo PRs run automatically. This gating can be removed if the updated Ledger test runner doesn't rely on private repos or secrets anymore.
- **Example:** A migration of existing tests to the new schema, exercising the full pipeline, can be seen here: https://github.com/manuelwedler/clear-signing-erc7730-registry/pull/2 — still in progress; not all tests have been ported to the new format yet.

## What's missing

- Ledger updating their test runner to consume the new `testsv2/` input format and emit the unified `results.json` output format.
- Migration of all existing test files to the new test file schema.
- Cleanup of legacy `tests/` artifacts once Ledger's runner has migrated and all test files are ported; the `testsv2/` folder can then be renamed back to `tests/`.


llbartekll:
Heads up — the upstream work needed to remove the fork dependency just merged on our side: https://github.com/llbartekll/clear-signing/pull/2 (now on `main`). That landed:

- registry-compatible `results.json` emission (Manuel's original work, attributed)
- our own `runner` identifier
- ENS / local name resolution split into two stub maps (`ensNames` + `addressNames`)
- `owner` threading on `DisplayEntry::Nested` so nested calldata frames carry the inner descriptor's owner

A few small follow-ups would let this PR drop the fork and tighten a couple of spec edges. Happy to push them as a PR to your `common-test-strategy` branch if easier — otherwise here they are inline:

---

### 1. `.github/actions/run-rust-tests/action.yml` — point at upstream

The fork + `.claude/worktrees/` workaround can go away entirely now that `cs-test` is on our `main`. Replace the action body with:

```yaml
runs:
  using: composite
  steps:
    - uses: dtolnay/rust-toolchain@stable
    - uses: Swatinem/rust-cache@v2

    - name: Install cs-test
      shell: bash
      run: |
        cargo install --git https://github.com/llbartekll/clear-signing \
          --branch main --locked cs-test

    - name: Run tests
      shell: bash
      env:
        TEST_FILE: ${{ inputs.test-file }}
        OUTPUT: ${{ inputs.output }}
      run: cs-test run "$TEST_FILE" --output "$OUTPUT"
```

That collapses the `git clone` + `rm -rf .claude` + `working-directory: rust-test-runner` steps to a single `cargo install --git` against upstream. Binary lands in `~/.cargo/bin` and is on `PATH`. Also lets you drop the `TODO: switch back to actions/checkout` comment.

### 2. Split ENS from local addresses (schema + aave fixture)

Our cs-test stub now routes `resolve_ens_name` and `resolve_local_name` to separate maps so descriptor `sources: ["ens"]` doesn't spuriously match a contact label and vice versa. Two coupled edits:

**`specs/erc7730-tests-v2.schema.json`** — add `ensNames` alongside `addressNames`:

```json
"ensNames": {
  "type": "object",
  "description": "Address → ENS name map. Separate from `addressNames` so descriptors specifying `sources: [\"ens\"]` are evaluated against ENS-sourced names only.",
  "additionalProperties": { "type": "string" }
}
```

**`registry/aave/testsv2/calldata-lpv2.tests.json`** — move `sosalkin.eth` (ENS) out of `addressNames` (local contact labels):

```diff
   "addressNames": {
-    "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2": "WETH",
-    "0xd20c9018a5097e922e9c0539aef389c871e76c3f": "sosalkin.eth"
+    "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2": "WETH"
+  },
+  "ensNames": {
+    "0xd20c9018a5097e922e9c0539aef389c871e76c3f": "sosalkin.eth"
   }
```

The aave field uses `sources: ["local", "ens"]` so it passes either way today — but fixtures that specify only `["ens"]` would silently pass the wrong runners under the current shared-map layout.

.github/actions/run-ledger-tests/action.yml
```diff
@@ -0,0 +1,180 @@
...
+outputs:
+  exit-code:
+    description: Exit code of the cs-tester run (0 = all tests passed / clear signed).
+    value: ${{ steps.run-tests.outputs.exit_code }}
+
+runs:
+  using: composite
+  steps:
+    - name: Checkout device-management-kit
+      uses: actions/checkout@v6.0.2
```
elasticroentgen: Use SHA hashes of the commit instead of the tag for all actions. Tags are mutable and can be switched to malicious commits. 

Do this for all action `uses` statements in the PR.

.github/actions/run-ledger-tests/action.yml
```diff
@@ -0,0 +1,180 @@
...
+      with:
+        repository: LedgerHQ/device-sdk-ts
+        ref: develop
```
elasticroentgen: similar issue here. Pulls in a mutable ref from an upstream repo. At least switch to a tag for deterministic CI results if the upstream can be trusted.

Similar cases across the workflow files. won't repeat

.github/actions/run-ledger-tests/action.yml
```diff
@@ -0,0 +1,180 @@
...
+        repository: LedgerHQ/coin-apps
+        ref: master
+        token: ${{ steps.generate-token.outputs.token }}
+        path: coin-apps
+        sparse-checkout: ${{ inputs.device }}/*/Ethereum/*.elf
+        sparse-checkout-cone-mode: false
+
+    - name: Pull Speculos Docker image
+      shell: bash
+      run: docker pull ghcr.io/ledgerhq/speculos
```
elasticroentgen: docker images points at `latest` - chance for non-deterministic CI results.
Pin to image SHA (to also prevent mutable image references) or at least a tag.

elasticroentgen (dismissed):
Please see the comments and apply the changes. Thanks!

manuelwedler:
@llbartekll I applied the change to the github action as you proposed but the CI is failing again for the claude submodule issue.

https://github.com/manuelwedler/clear-signing-erc7730-registry/actions/runs/26571333783/job/78278989862?pr=2

```

Run cargo install --git https://github.com/llbartekll/clear-signing \
    Updating git repository `[https://github.com/llbartekll/clear-signing`](https://github.com/llbartekll/clear-signing%60)
error: failed to update submodule `.claude/worktrees/competent-joliot-876b10`

Caused by:
  no URL configured for submodule '.claude/worktrees/competent-joliot-876b10'; class=Submodule (17)
Error: Process completed with exit code 101.
```

llbartekll:
@manuelwedler — confirmed, that's the same `.claude/worktrees/` cruft you hit on the original integration. My earlier claim that `cargo install --git` would sidestep it was wrong — cargo also tries to init submodules during the git fetch.

Cleanup PR queued up: https://github.com/llbartekll/clear-signing/pull/5. Three stale gitlinks removed from the index, `.claude/worktrees/` added to `.gitignore`. Once that merges to `main`, your `cargo install --git --branch main` invocation should work end-to-end.

manuelwedler:
@elasticroentgen thanks for checking. I fixed your comments in b843ac3cbeea91c3575490b0b6fa675da655f1d2

manuelwedler:
I migrated all existing test data to the new format and fixed the Sourcify implementation to fully match the expectations. Test results can be seen in these PRs:

https://github.com/manuelwedler/clear-signing-erc7730-registry/pull/2
https://github.com/manuelwedler/clear-signing-erc7730-registry/pull/3
https://github.com/manuelwedler/clear-signing-erc7730-registry/pull/4

There is only one case left failing on the Sourcify library because it is expected that it is not supported at the moment.

manuelwedler:
Update: I worked together with @llbartekll to fix any issues on the Rust library too. Both libraries run all tests successfully now (only one expected failure as stated above). See the latest runs here:
- https://github.com/manuelwedler/clear-signing-erc7730-registry/pull/5
- https://github.com/manuelwedler/clear-signing-erc7730-registry/pull/6
- https://github.com/manuelwedler/clear-signing-erc7730-registry/pull/7

hesterbruikman (approved):
Approving based on successful test runs against React and Rust library shared 

https://github.com/manuelwedler/clear-signing-erc7730-registry/pull/5
https://github.com/manuelwedler/clear-signing-erc7730-registry/pull/6
https://github.com/manuelwedler/clear-signing-erc7730-registry/pull/7

@elasticroentgen there are open change requests by you. Could you please give these another review to keep if blocking, dismiss if not?

Rafazzzz:
ab8c3b83c54ca8e9a84ccdff87c9def24982dab2

PR #2589, opened 2026-05-22
feat: add monad staking

ev4mrz:
- Add ERC-7730 V2 calldata descriptor for the Monad Staking precompile (chain 143, `0x0000000000000000000000000000000000001000`)

- Add 7 test cases covering: delegate, undelegate, withdraw, compound, claimRewards, changeCommission, addValidator


lcastillo-ledger (approved):

portdeveloper:
port from Monad Foundation devrel here. Nice work @ev4mrz, this is the first Monad-native entry in the registry.

I checked the descriptor against our official docs and it lines up:

- The target `0x0000000000000000000000000000000000001000` is correct. That is Monad's staking precompile at `0x1000` (see [Precompiles](https://docs.monad.xyz/developer-essentials/precompiles) and the [staking API reference](https://docs.monad.xyz/reference/staking/api)).
- The seven functions covered (delegate, undelegate, withdraw, compound, claimRewards, changeCommission, addValidator) all match the staking precompile's state-changing functions.

One thing for completeness: the precompile also exposes `externalReward` as a state-changing function (see the [staking API reference](https://docs.monad.xyz/reference/staking/api)). Not a blocker here, but worth a follow-up so the full staking surface is clear-signable.

On the red check: it is the "setup pull request" workflow's "Add labels" step (actions/labeler) hitting a transient HttpError, not anything in the descriptor. The schema and descriptor validation are green, the same setup step passes on current PRs, and this PR already carries the required `descriptors` label. Could a maintainer re-run the failed job and merge? It is already approved (thanks @lcastillo-ledger).

For context, Ledger is shipping native Monad staking in Ledger Live right now (delegations and undelegate landed in the last couple of weeks), so merging this makes that flow render human-readably end to end.

marcocastignoli (changes requested):
Thanks for the contribution! The descriptor itself is technically solid: all function signatures match the official Monad staking precompile interface (verified selectors against the decoded test transactions), and all 7 test transactions are real, successful mainnet transactions. A few changes are needed though:

**1. Native precompile / Sourcify**: `0x0000000000000000000000000000000000001000` has no bytecode on chain (`eth_getCode` returns `0x`), so it cannot be verified on Sourcify. We are currently thinking about how to handle native precompiles on the Sourcify side. No action needed from you on this point.

**2. Tests must be updated to the v2 format**: the test file uses the legacy schema (`erc7730-tests.schema.json` with `expectedTexts`) in `registry/monad/tests/`. CI requires `registry/monad/testsv2/calldata-MonadStaking.tests.json` using `erc7730-tests-v2.schema.json` (with a `descriptor` reference and `expected: { intent, fields: [{label, value}] }`). This is why the "Require testsv2 files" check fails, and the formatter test jobs are skipped entirely, so the tests never actually run.

**3. Fix `changeCommission` display**: the commission is 1e18 fixed-point (10% = 1e17 per the Monad docs), but the field uses `"format": "raw"`, so a user would see `100000000000000000` instead of a percentage. Please use `"format": "unit"` with `"params": { "base": "%", "decimals": 16 }`, which renders it as "10 %". Also worth asserting the rendered value in the test.

**4. Add an `externalReward` test**: the descriptor defines 8 formats but only 7 have tests. Please add a test case for `externalReward`.

**5. Consider adding `interpolatedIntent`**: the static one-word intents work, but this contract is a great candidate for interpolated intents, e.g. `"Delegate {@.value} to validator {#.validatorId}"`, `"Undelegate {#.amount} from validator {#.validatorId}"`, `"Change commission of validator {#.validatorId} to {#.commission}"`.

**6. Unused constants**: `metadata.constants` (`nativeTokenTicker`, `nativeDecimals`) are declared but never referenced anywhere in the descriptor. Please either use them or remove them.

*Posted with Claude Code*

PR #2590, opened 2026-05-25
feat: add rango diamond descriptors

RyukTheCoder:
Adds a new registry/rango/ entity covering Rango Diamond contract.


RyukTheCoder:
Hi @hesterbruikman, just bumping this PR as it's been open for a while and is ready for review. Let me know if there's anything else needed from my end! Thanks!

manuelwedler:
Hey there :)

RyukTheCoder:
> Hey there :) We've been recently adapting the testing CI to run tests against the TypeScript and Rust libraries at the same time. For this, we needed to change the test file format. Could you please do the following:
> 
> * Rebase this PR on latest master
> * Have a look at the updated [README](https://github.com/ethereum/clear-signing-erc7730-registry#reference-test-cases) for documentation on the new test format
> * Add test files in the v2 format
> * Remove the test files in the old format
> 
> Thank you! We will have a look at the PR afterwards.

Thanks for the guidance, @manuelwedler! I've made the requested changes:

- Rebased on the latest `master`.
- Migrated the tests to the v2 format. Added `registry/rango/testsv2/calldata-RangoDiamond.tests.json` using the new `erc7730-tests-v2.schema.json`. Each case now uses a structured expected block (intent / owner / fields) instead of expectedTexts, with a dataProvider.tokens block so token symbols resolve without network access.
- Removed the old-format file under `registry/rango/tests/`.
I also validated the fixture locally by building the `sourcifyeth/clear-signing-test-runner` (at the ref the CI pins) and running it against the new file, all 4 cases pass:

```
[pass] On-chain swap: 0.009985 ETH -> USDC on Ethereum (RangoSwapperFacet.onChainSwaps)
[pass] Across bridge-only: 0.0149775 ETH from Ethereum to Arbitrum (RangoAcrossFacet.acrossBridge)
[pass] AllBridge bridge-only: 9.985 USDT from Ethereum to Polygon (RangoAllBridgeFacet.allbridgeBridge)
[pass] Stargate V2 bridge-only: 9.985 USDT from Ethereum to Arbitrum (RangoStargateV2Facet.stargateV2Bridge)
```
done: 4 pass, 0 fail, 0 error, 0 skipped
Let me know if there's anything else you'd like me to adjust.

RyukTheCoder:
Hi @manuelwedler. All checks are green on [3bbebfe ](https://github.com/ethereum/clear-signing-erc7730-registry/pull/2590/changes/3bbebfe087eedd2eb53280187d6eece1fe278516)and the tests pass against both libs. Anything else needed from my side?

fbwoolf (changes requested):
The signature work here is genuinely careful, but the visibility choices undo most of it. One
blocking item, then the rest.

**1. A bridge's destination is never shown.** Every one of the 18 formats hides the destination
chain, and six of them hide the destination address too. Your own fixtures show the effect:

| fixture | what the signer sees |
| --- | --- |
| "AllBridge bridge-only: 9.985 USDT from Ethereum **to Polygon**" | `Bridge 9.985 USDT` |
| "Stargate V2 bridge-only: 9.985 USDT from Ethereum **to Arbitrum**" | `Bridge 9.985 USDT` |

Two bridges to different chains render as the same string. `stargateBridge`, `stargateV2Bridge` and
`allbridgeBridge` expose exactly one visible field, the amount. `stargateRequest.to` and
`stargateRequest.dstChainId` are both defined and both `visible: "never"`, so the recipient and the
chain are decoded and then withheld.

`docs/REVIEWING.md` puts this squarely in the reviewer's lane: "Value-bearing parameters, amounts,
spenders, recipients, deadlines, must never be hidden." For a cross-chain bridge the destination
chain belongs in that list ahead of almost everything else.

Also hidden in every format: `platformFee`, `affiliateFee`, `destinationExecutorFee`, and per-bridge
fees like `totalRelayFeeAmount`, `stgFee`, `totalAffiliateFee`. These are amounts taken from the
signer. The counts are stark, 394 of the 432 fields carry no `format` and nearly all are hidden:

| format | visible | hidden |
| --- | --- | --- |
| `stargateBridge` | 1 | 20 |
| `stargateV2Bridge` | 1 | 18 |
| `allbridgeBridge` | 1 | 15 |
| `acrossBridge` | 2 | 18 |
| `stargateSwapAndBridge` | 2 | 29 |
| `onChainSwaps` | 3 | 17 |

**2. The linter validated nothing, and Sourcify cannot fix that here.** All 18 selectors are reported
as "not found in reference ABI". That is not a stale descriptor: the reference ABI resolves to the
EIP-2535 diamond's own fallback ABI, so no facet selector can ever appear in it. Verifying more
chains on Sourcify does not change this, which makes REVIEWING.md's usual remedy inapplicable.

I checked it the way that does work, `facetAddress(bytes4)` on the mainnet diamond, then the facet
ABI from Sourcify. All eight facets are verified with exact matches, and **all 16 live selectors match
their facet ABI exactly, including every nested tuple component name at every depth**. On a 2566-line
descriptor with tuples nested three deep that is a good result and worth saying.

**3. Two formats target a facet that is not deployed anywhere.** `0x401c2eb0` (`acrossBridge`) and
`0xbdd9c675` (`acrossSwapAndBridge`), the two carrying `decimalsAdjustedTotalRelayFeePct`, resolve to
no facet on any of the 25 chains I could reach, including all the major ones. Your fixture calls one
of them "new facet ABI", so I take it this is deliberate forward coverage. Worth confirming, and
worth knowing they are inert until that facet ships.

**4. Sourcify.** 19 of the 44 deployments were verified. I submitted the rest to
`/v2/verify/similarity` and 20 more now verify. Three cannot be done:

| chain | address | why |
| --- | --- | --- |
| 324 ZKsync Era, 2741 Abstract | `0x13598FD0…` | `no_similar_match_found`; Etherscan import is unsupported for 324 and 2741 is not verified there |
| 128 HECO, 43288 Boba Avax, 56288 Boba BNB | `0x69460570…` / `0xd9BdD77E…` | not supported by Sourcify at all |

**5. Two recipients render differently from the other fourteen.**
`satelliteBridge.request.receiver` and `satelliteSwapAndBridge.bridgeRequest.receiver` have no
`format` and no `params`, while every other recipient uses `addressName` with ENS resolution. Looks
unintended.

**6. Test coverage.** 5 cases for 18 formats, and all 5 are bridge-only or swap-only. No
`*SwapAndBridge` path is exercised, and those are the largest formats in the file (up to 32 fields).

Advisory, from the linter: nine labels exceed 20 characters and seven interpolated intents exceed 30,
so both will truncate on Ledger devices.

RyukTheCoder:
> The signature work here is genuinely careful, but the visibility choices undo most of it. One blocking item, then the rest.
> 
> **1. A bridge's destination is never shown.** Every one of the 18 formats hides the destination chain, and six of them hide the destination address too. Your own fixtures show the effect:
> 
> fixture	what the signer sees
> "AllBridge bridge-only: 9.985 USDT from Ethereum **to Polygon**"	`Bridge 9.985 USDT`
> "Stargate V2 bridge-only: 9.985 USDT from Ethereum **to Arbitrum**"	`Bridge 9.985 USDT`
> Two bridges to different chains render as the same string. `stargateBridge`, `stargateV2Bridge` and `allbridgeBridge` expose exactly one visible field, the amount. `stargateRequest.to` and `stargateRequest.dstChainId` are both defined and both `visible: "never"`, so the recipient and the chain are decoded and then withheld.
> 
> `docs/REVIEWING.md` puts this squarely in the reviewer's lane: "Value-bearing parameters, amounts, spenders, recipients, deadlines, must never be hidden." For a cross-chain bridge the destination chain belongs in that list ahead of almost everything else.
> 
> Also hidden in every format: `platformFee`, `affiliateFee`, `destinationExecutorFee`, and per-bridge fees like `totalRelayFeeAmount`, `stgFee`, `totalAffiliateFee`. These are amounts taken from the signer. The counts are stark, 394 of the 432 fields carry no `format` and nearly all are hidden:
> 
> format	visible	hidden
> `stargateBridge`	1	20
> `stargateV2Bridge`	1	18
> `allbridgeBridge`	1	15
> `acrossBridge`	2	18
> `stargateSwapAndBridge`	2	29
> `onChainSwaps`	3	17
> **2. The linter validated nothing, and Sourcify cannot fix that here.** All 18 selectors are reported as "not found in reference ABI". That is not a stale descriptor: the reference ABI resolves to the EIP-2535 diamond's own fallback ABI, so no facet selector can ever appear in it. Verifying more chains on Sourcify does not change this, which makes REVIEWING.md's usual remedy inapplicable.
> 
> I checked it the way that does work, `facetAddress(bytes4)` on the mainnet diamond, then the facet ABI from Sourcify. All eight facets are verified with exact matches, and **all 16 live selectors match their facet ABI exactly, including every nested tuple component name at every depth**. On a 2566-line descriptor with tuples nested three deep that is a good result and worth saying.
> 
> **3. Two formats target a facet that is not deployed anywhere.** `0x401c2eb0` (`acrossBridge`) and `0xbdd9c675` (`acrossSwapAndBridge`), the two carrying `decimalsAdjustedTotalRelayFeePct`, resolve to no facet on any of the 25 chains I could reach, including all the major ones. Your fixture calls one of them "new facet ABI", so I take it this is deliberate forward coverage. Worth confirming, and worth knowing they are inert until that facet ships.
> 
> **4. Sourcify.** 19 of the 44 deployments were verified. I submitted the rest to `/v2/verify/similarity` and 20 more now verify. Three cannot be done:
> 
> chain	address	why
> 324 ZKsync Era, 2741 Abstract	`0x13598FD0…`	`no_similar_match_found`; Etherscan import is unsupported for 324 and 2741 is not verified there
> 128 HECO, 43288 Boba Avax, 56288 Boba BNB	`0x69460570…` / `0xd9BdD77E…`	not supported by Sourcify at all
> **5. Two recipients render differently from the other fourteen.** `satelliteBridge.request.receiver` and `satelliteSwapAndBridge.bridgeRequest.receiver` have no `format` and no `params`, while every other recipient uses `addressName` with ENS resolution. Looks unintended.
> 
> **6. Test coverage.** 5 cases for 18 formats, and all 5 are bridge-only or swap-only. No `*SwapAndBridge` path is exercised, and those are the largest formats in the file (up to 32 fields).
> 
> Advisory, from the linter: nine labels exceed 20 characters and seven interpolated intents exceed 30, so both will truncate on Ledger devices.

Thanks for the thorough review @fbwoolf — especially validating the selectors via `facetAddress(bytes4)` and pushing the extra deployments through Sourcify similarity-verify. Pushed changes addressing everything:

**1 (blocking) — destination + fees.** All 18 formats now surface the destination, and fees are shown:
- **Destination chain** in every format. EVM chain-id fields (Across, Generic, Stargate V2, Symbiosis) use the `chainId` format → EIP-155 name (`Arbitrum One`, `Polygon Mainnet`). AllBridge and Stargate V1 carry *proprietary* ids (Rango's interface documents AllBridge's as "specific chain id for allbridge, not general evm chain id"; Stargate V1's is a LayerZero id) — `chainId` would resolve a wrong name (AllBridge `5` → "Goerli"), so they show as `raw` labeled **Destination Chain ID**. The two USDT bridges that used to both render `Bridge 9.985 USDT` are now distinct.
- **The 6 previously-hidden recipients** (AllBridge / Stargate V1 / Stargate V2, bridge + swap-and-bridge) are now visible. They're `bytes`/`bytes32`, so `raw` hex rather than `addressName`.
- **Fees.** On Bridge formats `platformFee`/`affiliateFee`/`destinationExecutorFee`/`totalAffiliateFee`/`totalRelayFeeAmount` render as `tokenAmount` in the bridge token (denomination is fixed by `RangoBridgeRequest.token`). On SwapAndBridge/onChainSwaps the same fees are shown `raw`: those structs carry a `feeFromInputToken` flag so the fee token is chosen at runtime — a static `tokenPath` would show wrong decimals half the time, so raw rather than mislead. Native/protocol fees (`stgFee`, `nativeFee`, `transferFee`, `feeTokenAmount`, `relayerGas`) are `raw`.

**2 — reference-ABI selector warnings.** Agreed, unavoidable with the EIP-2535 fallback ABI; thanks for verifying them the correct way. No change.

**3 — `0x401c2eb0` / `0xbdd9c675`.** Confirmed intentional forward coverage — the `decimalsAdjustedTotalRelayFeePct` Across facet isn't deployed yet; the formats are inert until it ships.

**4 — Sourcify.** Thanks for the +20. The last 5 are known zk/unsupported-chain limits (ZKsync Era, Abstract, HECO, Boba Avax, Boba BNB) — nothing further available our side.

**5 — `satellite*.receiver` string.** By necessity: Satellite (Axelar) takes the destination address as a `string`, not `address`, so ENS/`addressName` can't apply.

**6 — coverage.** Added 3 SwapAndBridge fixtures (`acrossSwapAndBridge`, `stargateV2SwapAndBridge`, `genericSwapAndBridge`) covering the largest formats plus the new destination/recipient/fee fields. Suite is now 8 cases; all pass against `@ethereum-sourcify/clear-signing@0.2.2` (ran the pinned runner locally).

Also trimmed the surfaced labels to ≤20 chars per the Ledger advisory — `erc7730 lint` is at 0 errors, and the label-length advisory dropped from 9 to 6 (the remaining 6 are hidden internal fields). The intent/selector warnings are unchanged (the latter being the diamond fallback-ABI artifact from point 2).

RyukTheCoder:
Heads up: CI is red on one case, and I think it is an implementation divergence rather than a descriptor issue.

genericSwapAndBridge to Polygon renders Destination Chain as Polygon Mainnet under @ethereum-sourcify/clear-signing@0.2.2 and Polygon under the pinned llbartekll/clear-signing. Chain 42161 agrees in both, so the other five chainId fixtures pass and only chain 137 fails. I had verified locally against the Sourcify runner only, which is why I missed it.

This descriptor looks to be the only one in the registry using format: "chainId", so I think this is the first time the dual runner has exercised it. Happy to go whichever way you prefer: leave it as is while the two libs are aligned, re point that fixture at a chain both name identically, or fall back to format: "raw" for the chain fields the way calldata-LIFIDiamond.json does. Say the word and I will push it.
@fbwoolf

fbwoolf:
Your first option: leave the descriptor and the fixture as they are. The fixture is already the spec-conformant side, so "aligned" here means the Rust runner changes, not the expectation.

`specs/erc-7730.md` defines the format as "value is converted to a Blockchain name using EIP-155 reference values", and its worked example is `Ethereum Mainnet` for value 1. That string is the `name` field from the EIP-155 chain list at chainid.network, which gives 1 as `Ethereum Mainnet`, 137 as `Polygon Mainnet`, and 42161 as `Arbitrum One`. That also explains why only chain 137 fails: `Arbitrum One` carries no "Mainnet" suffix, so both libraries land on the same string, while `Polygon Mainnet` does and the Rust runner shortens it. Worth reporting against `llbartekll/clear-signing`, which the registry pins at `10605ba`.

The other two options both cost more than they save. Falling back to `raw` puts `137` on the device, which is the illegible destination this review was about in the first place, so it trades a runner bug for the original defect. Re-pointing the fixture at a chain both libraries name identically goes green while real users bridging to Polygon still see different strings in different wallets, since both of these are wallet-facing implementations rather than test harnesses.

The LIFI precedent does not hold either: `calldata-LIFIDiamond.json` has no chain-related field at all, `raw` or otherwise, so there is nothing there to follow.

On the wider change, using `chainId` only where the value really is an EIP-155 id, and leaving Stargate v1, AllBridge and Satellite as raw because theirs are LayerZero domain ids, AllBridge ids and Axelar chain names, is the right split. One leftover from the earlier review while you are pushing: `satelliteBridge`'s `request.receiver` still has no `format`, so it renders as a bare address while the other recipients resolve names.

RyukTheCoder:
Thanks @fbwoolf, that settles it. Going with option one.

**Polygon case, left as is.** The fixture keeps `Polygon Mainnet` for chain 137, which is the EIP-155 `name` value that the spec's own `Ethereum Mainnet` example points to. Neither the descriptor nor the expectation has been touched. I am subscribed to llbartekll/clear-signing#14 and will ping here if the reading gets pushed back on upstream, since that would change the calculus on this PR.

For the record, the Rust job reports exactly one assertion. Nothing else in the suite diverges, and every other check on the PR is green:

```
Swap and Bridge: swap 100 USDC then generic-bridge to Polygon (RangoGenericBridgeFacet.genericSwapAndBridge) — field value at [4] "Destination Chain": expected "Polygon Mainnet", got "Polygon"
```

**Satellite recipient.** Fixed in e72927e, with one adjustment to what you suggested. `addressName` does not apply to this field: both Satellite receivers are typed `string` in the ABI rather than `address`, since Axelar carries destinations that are not EVM addresses. The spec scopes `addressName` to the `address` solidity type and defines `raw` as the only format for `string`, so both now carry `format: "raw"`. That also matches how the AllBridge (`bytes32`) and Stargate (`bytes`) recipients are already handled in this descriptor.

I applied the same to `satelliteSwapAndBridge`'s `bridgeRequest.receiver`, which had the identical gap.

One note while you review: the fixture has no Satellite case, so neither runner exercises either of these fields. Happy to add one if you want the coverage.

PR #2592, opened 2026-05-25
Add Privacy Pools clear signing descriptors

patcito:
Adds clear signing support for Privacy Pools wallet-facing flows.

Covers entrypoint deposits, pool deposits, ragequit exits, pool withdrawals, wind down, and the seed derivation EIP-712 signature used by the app.

Withdrawals usually go through the relayer API, but the pool transaction format is included so the signing path is covered if it appears in a wallet.


manuelwedler:
Hey there :)

beepboop2025:
I ran this with a tool I've been building: github.com/beepboop2025/lucent.

The entrypoint address is an EIP-1967 proxy; this review runs against the
implementation ABI (0x15e355024de1cdc74addea7ebdf98418ba5b1a2c).

In order of importance, I want to point out three issues:

(1) both deposit entries embed parameter names in their keys, so as submitted
neither ever matches a wallet lookup - the descriptor currently renders
nothing for users; canonicalizing the two keys is a one-line fix each.

(2) relay(), the user-facing withdrawal path, has no entry at all - worth
covering, it is exactly the call users sign.

(3) the admin/role/upgrade functions being uncovered is often deliberate;
treat those findings as informational, but upgradeToAndCall is why the tool's
agent gate returns block here. Same non-canonical-key defect appears in all
three PRs I reviewed this week - possibly a shared generator/template issue
upstream. 

# ERC-7730 descriptor review

## calldata-PrivacyPoolEntrypoint

- deployment: `eip155:1 0x6818809EefCe719E480a7526D76bD3e561526b46`
- deployment: `eip155:10 0x44192215FEd782896BE2CE24E0Bfbf0BF825d15E`
- deployment: `eip155:56 0x44192215FEd782896BE2CE24E0Bfbf0BF825d15E`
- deployment: `eip155:42161 0x44192215FEd782896BE2CE24E0Bfbf0BF825d15E`
- deployment: `eip155:11155111 0x34A2068192b1297f2a7f85D7D8CdE66F8F0921cB`
- deployment: `eip155:11155420 0x54aCA0D27500669FA37867233e05423701f11ba1`
- signable functions: 14, covered: 0

**Verdict: block** - 2 CRITICAL danger primitive(s) — a clear screen cannot make this safe to sign blind

### Lint: pass
- ABI validation: skipped (no ETHERSCAN_API_KEY)

### Screen audit: grade F (0/100)
- HIGH `deposit`: format entry keyed non-canonically as 'deposit(address _asset, uint256 _value, uint256 _precommitment)' - wallets match the canonical signature, so this entry never renders (blind-signs as submitted)
- HIGH `deposit`: format entry keyed non-canonically as 'deposit(uint256 _precommitment)' - wallets match the canonical signature, so this entry never renders (blind-signs as submitted)
- HIGH `grantRole`: signable function has no descriptor entry (blind-signs)
- HIGH `initialize`: signable function has no descriptor entry (blind-signs)
- HIGH `registerPool`: signable function has no descriptor entry (blind-signs)
- HIGH `relay`: signable function has no descriptor entry (blind-signs)
- HIGH `removePool`: signable function has no descriptor entry (blind-signs)
- HIGH `renounceRole`: signable function has no descriptor entry (blind-signs)
- HIGH `revokeRole`: signable function has no descriptor entry (blind-signs)
- HIGH `updatePoolConfiguration`: signable function has no descriptor entry (blind-signs)
- HIGH `updateRoot`: signable function has no descriptor entry (blind-signs)
- HIGH `upgradeToAndCall`: signable function has no descriptor entry (blind-signs)
- HIGH `windDownPool`: signable function has no descriptor entry (blind-signs)
- HIGH `withdrawFees`: signable function has no descriptor entry (blind-signs)

### Comprehension: grade F (0/100), worst tier CRITICAL
- CRITICAL `grantRole`
  - screen: You change the control or code of the contract (grantRole), affecting everyone who uses it.
  - why: changes WHO controls the contract or WHAT code runs — a consequence far beyond moving your own funds
- CRITICAL `upgradeToAndCall`
  - screen: You change the control or code of the contract (upgradeToAndCall), affecting everyone who uses it.
  - why: changes WHO controls the contract or WHAT code runs — a consequence far beyond moving your own funds
- MEDIUM `deposit`
  - screen: You call deposit — the screen does not explain what this does.
  - why: unclassified consequence AND no intent on the screen — the human has nothing to understand the action from (an unexplained screen is worse than none: it invites blind approval)
- MEDIUM `deposit`
  - screen: You send {amount} to a recipient now.
  - why: moves value out of your wallet now — verify the recipient is who you expect
- MEDIUM `initialize`
  - screen: You call initialize — the screen does not explain what this does.
  - why: unclassified consequence AND no intent on the screen — the human has nothing to understand the action from (an unexplained screen is worse than none: it invites blind approval)
- MEDIUM `registerPool`
  - screen: You call registerPool — the screen does not explain what this does.
  - why: unclassified consequence AND no intent on the screen — the human has nothing to understand the action from (an unexplained screen is worse than none: it invites blind approval)
- MEDIUM `relay`
  - screen: You call relay — the screen does not explain what this does.
  - why: unclassified consequence AND no intent on the screen — the human has nothing to understand the action from (an unexplained screen is worse than none: it invites blind approval)
- MEDIUM `removePool`
  - screen: You call removePool — the screen does not explain what this does.
  - why: unclassified consequence AND no intent on the screen — the human has nothing to understand the action from (an unexplained screen is worse than none: it invites blind approval)
- MEDIUM `renounceRole`
  - screen: You call renounceRole — the screen does not explain what this does.
  - why: unclassified consequence AND no intent on the screen — the human has nothing to understand the action from (an unexplained screen is worse than none: it invites blind approval)
- MEDIUM `revokeRole`
  - screen: You call revokeRole — the screen does not explain what this does.
  - why: unclassified consequence AND no intent on the screen — the human has nothing to understand the action from (an unexplained screen is worse than none: it invites blind approval)
- MEDIUM `updatePoolConfiguration`
  - screen: You call updatePoolConfiguration — the screen does not explain what this does.
  - why: unclassified consequence AND no intent on the screen — the human has nothing to understand the action from (an unexplained screen is worse than none: it invites blind approval)
- MEDIUM `updateRoot`
  - screen: You call updateRoot — the screen does not explain what this does.
  - why: unclassified consequence AND no intent on the screen — the human has nothing to understand the action from (an unexplained screen is worse than none: it invites blind approval)
- MEDIUM `windDownPool`
  - screen: You call windDownPool — the screen does not explain what this does.
  - why: unclassified consequence AND no intent on the screen — the human has nothing to understand the action from (an unexplained screen is worse than none: it invites blind approval)
- MEDIUM `withdrawFees`
  - screen: You call withdrawFees — the screen does not explain what this does.
  - why: unclassified consequence AND no intent on the screen — the human has nothing to understand the action from (an unexplained screen is worse than none: it invites blind approval)

### Danger surface
- CRITICAL `relay`: arbitrary-external-call
  - declares an unbounded external call (call-family name or a target-address + calldata-blob signature) — the signer cannot bound its effect by reading the screen; the canonical wallet-drainer primitive. A clear sentence cannot make it safe.
- CRITICAL `upgradeToAndCall`: upgrade-and-execute
  - swaps the contract's code AND runs a call in one signature — the single most powerful action a proxy exposes.
- HIGH `grantRole`: authority-transfer
  - changes who controls the contract (ownership/admin/role) — affects every user, not just the signer's funds.
- HIGH `renounceRole`: authority-transfer
  - changes who controls the contract (ownership/admin/role) — affects every user, not just the signer's funds.
- MEDIUM `withdrawFees`: value-sweep-to-arbitrary-address
  - moves held value to a caller-supplied address — verify the destination is not attacker-controlled.

### Semantic verification
- skipped: no test vectors at tests/calldata-PrivacyPoolEntrypoint.tests.json

---

Produced by Lucent (audit + comprehension + danger + semantic verification
over ERC-7730 descriptors). Reproduce any section from the repo root:

```
make audit DESC=<descriptor.json>
make comprehend DESC=<descriptor.json>
make danger DESC=<descriptor.json>
make semverify DESC=<descriptor.json>   # needs ETHERSCAN_API_KEY
```

manuelwedler:

patcito:
@manuelwedler done

registry/privacy-pools/calldata-PrivacyPool.json
```diff
@@ -0,0 +1,235 @@
...
+                "local",
+                "ens"
+              ]
+            },
+            "visible": "always"
+          },
+          {
+            "path": "_value",
+            "label": "Amount",
+            "format": "raw",
```
kuzdogan: Is this supposed to be a `raw` value really? It sounds like a token amount?

registry/privacy-pools/calldata-PrivacyPoolEntrypoint.json
```diff
@@ -0,0 +1,87 @@
...
+          {
+            "path": "_precommitment",
+            "label": "Precommitment",
+            "visible": "never"
+          }
+        ]
+      },
+      "deposit(address _asset,uint256 _value,uint256 _precommitment)": {
+        "intent": "Deposit token",
+        "interpolatedIntent": "Deposit {_value}",
```
kuzdogan, line 61: Here the `interpolatedIntent` is used correctly

registry/privacy-pools/calldata-PrivacyPool.json
```diff
@@ -0,0 +1,235 @@
...
+    "contractName": "Privacy Pool",
+    "info": {
+      "url": "https://0xbow.io"
+    }
+  },
+  "display": {
+    "formats": {
+      "deposit(address _depositor,uint256 _value,uint256 _precommitmentHash)": {
+        "intent": "Deposit into pool",
+        "interpolatedIntent": "Deposit into pool",
```
kuzdogan:
```suggestion
        "interpolatedIntent": "Deposit {_value} and {@.value} into pool from {_depositor} ",
```

The interpolated intent should look like this for all cases

kuzdogan (changes requested):
Please double check the interpolatedIntents, they are not used correctly. Also check other commetns

registry/privacy-pools/calldata-PrivacyPool.json
patcito: You're right, it's an amount and not opaque bytes. That's why the test expectation reads `100000000000000000`. Changed to `tokenAmount`.

I can't bind a ticker to it. This descriptor covers 25 pool addresses across 6 chains and each pool holds a different asset. `ASSET()` on the 14 mainnet pools returns 14 distinct tokens (USDC, USDT, DAI, WBTC, wstETH, USDS, sUSDe, and one native-ETH pool). `deposit(address,uint256,uint256)` takes no asset argument, so `tokenPath` has nothing to point at, and a constant `token` would be right for exactly one of the 25.

Per the format table in the spec, an unbound `tokenAmount` displays the raw value with an "Unknown token" warning. That matches what's true here: it is an amount, of an asset this descriptor cannot identify.

Real tickers would mean one descriptor per pool. That's the only way to get "1 USDC" rendering, so happy to split it that way if you'd prefer. It's a much larger change so I didn't assume.

registry/privacy-pools/calldata-PrivacyPoolEntrypoint.json
patcito: Thanks, used this as the reference for the rest.

The sibling `deposit(uint256 _precommitment)` above it was still static, so it's now `Deposit {@.value}`. Its `@.value` field was already `visible: "always"`.

registry/privacy-pools/calldata-PrivacyPool.json
patcito: Applied, with one prerequisite and one deviation.

The prerequisite: `@.value` in this format was `visible: "optional"`, and the spec says *"Interpolated intents MUST only refer to paths that are noted as always visible"*. So interpolating `{@.value}` needed it promoted to `always` first, which I've done. It's right on its own terms too. `docs/REVIEWING.md` says value-bearing parameters must never be hidden, and this is native value leaving the signer's wallet.

The deviation: I dropped the trailing space before the closing quote. It looked accidental.

Same treatment everywhere a format has an always-visible field to interpolate:

- `PrivacyPool.withdraw` → `Withdraw from pool via {_withdrawal.processooor}`
- `Entrypoint.deposit(uint256)` → `Deposit {@.value}`
- `DeriveSeed` → `Derive seed for {context}`

Two are deliberately left as plain strings. `windDown()` takes no arguments at all. `ragequit`'s only non-hidden field is the proof's public signals, which are `optional` and meaningless to a signer, so interpolating there would mean promoting proof internals to always-visible. Say the word if you'd rather they carried something.

Test expectations updated to the rendered strings. One caveat: I couldn't run the harness locally, since this repo ships only `generate-index`. So the unbound-`tokenAmount` value is written as the spec's stated fallback, the raw digits. If the harness renders the unknown-token case differently, that expectation needs a nudge.

registry/privacy-pools/calldata-PrivacyPool.json
patcito: Following up on my caveat above: CI has run the harness and all three descriptors pass, Rust and Sourcify tests plus schema, descriptor and index validation. So the unbound-`tokenAmount` expectation is confirmed, it does render the raw digits like the spec's format table describes.

PR #2595, opened 2026-05-29
Add Zama confidential wrappers

melanciani:


registry/zama/calldata-ConfidentialWrapper.json
```diff
@@ -0,0 +1,187 @@
...
+      "receiver": { "label": "Receiver", "format": "addressName", "params": { "types": ["eoa", "contract"] } },
+      "holder": { "label": "Holder", "format": "addressName", "params": { "types": ["eoa", "contract"] } },
+      "inputProof": { "label": "Encryption proof", "format": "raw" },
+      "decryptionProof": { "label": "Decryption proof", "format": "raw" },
+      "callbackData": { "label": "Callback data", "format": "raw" }
+    },
+    "formats": {
+      "wrap(address to, uint256 amount)": {
+        "$id": "wrap",
+        "intent": "Shield",
```
melanciani, line 46: should the intend be about (un)wrap or (un)shield ?

registry/zama/calldata-ConfidentialWrapper.json
```diff
@@ -0,0 +1,187 @@
...
+        ]
+      },
+      "unwrap(address from, address to, bytes32 encryptedAmount, bytes inputProof)": {
+        "$id": "unwrapWithProof",
+        "intent": "Request to unshield",
+        "fields": [
+          { "path": "from", "$ref": "$.display.definitions.holder" },
+          { "path": "to", "$ref": "$.display.definitions.receiver" },
+          { "path": "encryptedAmount", "$ref": "$.display.definitions.encryptedAmount" },
+          { "path": "inputProof", "$ref": "$.display.definitions.inputProof", "visible": "never" }
```
melanciani, line 71: should wallet display input proof ?

registry/zama/calldata-ConfidentialWrapper.json
```diff
@@ -0,0 +1,187 @@
...
+        ]
+      },
+      "finalizeUnwrap(bytes32 unwrapRequestId, uint64 unwrapAmountCleartext, bytes decryptionProof)": {
+        "$id": "finalizeUnwrap",
+        "intent": "Finalize unshield",
+        "fields": [
+          { "path": "unwrapAmountCleartext", "label": "Amount", "format": "tokenAmount", "params": { "tokenPath": "@.to" } },
+          { "path": "unwrapRequestId", "label": "Request ID", "format": "raw" },
+          { "path": "decryptionProof", "$ref": "$.display.definitions.decryptionProof", "visible": "never" }
```
melanciani, line 81: should wallet display decryption proof ?

registry/zama/calldata-ConfidentialWrapper.json
```diff
@@ -0,0 +1,187 @@
...
+      "inputProof": { "label": "Encryption proof", "format": "raw" },
+      "decryptionProof": { "label": "Decryption proof", "format": "raw" },
+      "callbackData": { "label": "Callback data", "format": "raw" }
+    },
+    "formats": {
+      "wrap(address to, uint256 amount)": {
+        "$id": "wrap",
+        "intent": "Shield",
+        "fields": [
+          { "path": "amount", "label": "Amount", "format": "raw" },
```
melanciani: note: wrap amounts are in “raw” (instead of "tokenAmount") because we can’t access the underlying token’s decimals using eip7730 semantics

registry/zama/calldata-ConfidentialWrapper.json
```diff
@@ -0,0 +1,174 @@
...
+      "holder": { "label": "Holder", "format": "addressName", "params": { "types": ["eoa", "contract"] } },
+      "inputProof": { "label": "Encryption proof", "format": "raw" },
+      "decryptionProof": { "label": "Decryption proof", "format": "raw" },
+      "callbackData": { "label": "Callback data", "format": "raw" }
+    },
+    "formats": {
+      "wrap(address to, uint256 amount)": {
+        "$id": "wrap",
+        "intent": "Shield",
+        "fields": [{ "path": "amount", "label": "Amount", "format": "raw" }, { "path": "to", "$ref": "$.display.definitions.receiver" }]
```
melanciani: note: wrap amounts are in “raw” (instead of "tokenAmount") because we can’t access the underlying token’s decimals using eip7730 semantics

specs/erc7730-v2.schema.json
```diff
@@ -563,6 +563,9 @@
                 {
                     "not": { "required": [ "path", "value" ] }
                 },
+                {
+                    "not": { "required": [ "$ref" ] }
```
melanciani: after some fight with claude and the CI (failures [as such](https://github.com/ethereum/clear-signing-erc7730-registry/actions/runs/26659311274/job/78577504285)), I think I've found a bug in the current way the schema is written. Here a summary from my discussions with claude cc @lcastillo-ledger : 

 ## TL;DR

The v2 schema declares **draft-07** (line 2), but `field` uses `unevaluatedProperties: false` (a draft-2019-09+ keyword) to reject stray keys. Draft-07 ignores that keyword, so `field` accepts an extra `$ref` and overlaps the `reference` branch → any bare `{ "path": ..., "$ref": ... }` field matches two `oneOf` branches and fails validation. Fix: either declare 2020-12, or add `{ "not": { "required": ["$ref"] } }` to `field`.

---

## Details

**Symptom.** `check-jsonschema` rejects bare references like `{ "path": "amount", "$ref": "$.display.definitions.x" }` with *"valid under each of `#/$display/reference`, `#/$format/field`"* (matches 2 of the `fields[]` `oneOf` branches).

**Cause.** Dialect mismatch: `field`'s `unevaluatedProperties: false` (line ~640) is meant to reject unknown keys, but it only exists from draft 2019-09 — under the declared draft-07 it's silently ignored, so `field` tolerates `$ref` and overlaps `reference`. The same mismatch overlaps `field`/`fieldGroup` for `{ "path": ..., "fields": [...] }` groups (e.g. `registry/paraswap/calldata-AugustusSwapper-v6.2.json` currently fails too).

**Why latent.** Only bites refs with no other distinguishing key. Refs that also carry `params` (e.g. `registry/kyberswap/calldata-MetaAggregationRouterV2.json`) accidentally fail the `field` branch and pass; bare refs (`receiver`, `holder`, proofs) don't. `ercs/calldata-erc721-nfts.json` is affected but isn't CI-validated.

**Fix options:**

1. **Root cause:** set line 2 to `https://json-schema.org/draft/2020-12/schema` → activates the existing `unevaluatedProperties`, fixes both overlaps, keeps strictness. ⚠️ changes whole-file dialect (e.g. `$ref`-with-siblings, line 546) — needs full-registry re-validation.
2. **Targeted, draft-07-safe (applied):** add `{ "not": { "required": ["$ref"] } }` to `field`'s `allOf`. Verified green on the affected file + `kyberswap` / `erc721` / `okx` / `1inch` / `lido` / `threshold`; no regression. Doesn't cover the `fieldGroup` overlap (would need an analogous `{ "not": { "required": ["fields"] } }`).
3. **Permissive:** `oneOf` → `anyOf` (line 419) — one word, fixes both, but weakens validation.

**Recommendation:** option 1 long-term (the dialect is just mis-declared); option 2 short-term/low-risk.

registry/zama/calldata-ConfidentialWrapper.json
melanciani: let's keep shield

specs/erc7730-v2.schema.json
melanciani: I opened a PR against ERC repo here: https://github.com/ethereum/ERCs/pull/1822

melanciani: I updated https://github.com/ethereum/ERCs/pull/1822 to consider solution 1 instead

melanciani: note: 
- I've made Claude run a check on all current schemas in the registry repo and only a single schema fails, but **because of a pre-existing unrelated issue** : `registry/paraswap/calldata-AugustusSwapper-v5.json`
- this can be checked by running `check-jsonschema --schemafile specs/erc7730-v2.schema.json registry/paraswap/calldata-AugustusSwapper-v5.json` in main currently

melanciani: updating the schema has been moved to https://github.com/ethereum/clear-signing-erc7730-registry/pull/2624 , it should be merged separately (and first)

hilljessica8787-dot (approved):

hilljessica8787-dot (approved):

manuelwedler:
Hey there :)

manuelwedler:
Note that tests will likely fail because at least the TypeScript library does not support encryption yet (I assume the Rust library too).

melanciani:
@manuelwedler I've updated the test for the v2 version, ran them locally and confirm there is some missing logic in the ts library (and indeed probably rust), do you confirm they come from https://github.com/sourcifyeth/clear-signing/ and https://github.com/llbartekll/clear-signing ? am i expected to open PRs there ?

hilljessica8787-dot (approved):

manuelwedler:
@melanciani I don't think you are expected to open PRs. We should get the libraries to a state where the full 7730 spec is implemented. So I just created a draft PR on the sourcify library: https://github.com/sourcifyeth/clear-signing/pull/56. You can take a look at it if the implementation makes sense like this (I didn't verify what is written in `DECRYPTION.md` myself yet though).

I'm undecided if we need to get this implemented on both libraries before this can be merged, as it's a very specific feature. Nevertheless, you can create a PR on `llbartekll/clear-signing` yourself if you feel like it.

melanciani:
thanks a lot @manuelwedler , I've reviewed your PR (just made a minor comment) + opened an equivalent of your PR in the rust lib: https://github.com/llbartekll/clear-signing/pull/10 ! 

imo having support for the encryption scheme is a great benefit to the libs so would be great to try to integrate them both !

hilljessica8787-dot (approved):

melanciani:
hey @manuelwedler , is there a way to run the ts tests now that https://github.com/sourcifyeth/clear-signing/pull/56 has been merged and v0.2.2 has been released ? 

also: are failing tests from the rust lib a blocker to merge this PR ?

melanciani:
I verified locally that clear-signing-test-runner's `dae3cdabd0eab26173d7f7a31a2ca7e75bf07daf` commit fixes tests in this PR 

could you re-run tests with it @manuelwedler ? until https://github.com/ethereum/clear-signing-erc7730-registry/pull/2866 gets merged. Also, are the Rust tests a blocker for merging this PR ? I still have no feedback for https://github.com/llbartekll/clear-signing/pull/10. Thanks !

melanciani:
hey again @manuelwedler , the rust lib now supports the encryption feature ! I've updated it in this PR so we can already re-run the tests. I've also opened this PR to update it in a dedicated commit: https://github.com/ethereum/clear-signing-erc7730-registry/pull/2888 (took your PR as an example for its description). Once merged, I'll rebase this PR !

melanciani:
rebased to include new Rust lib test runner: https://github.com/ethereum/clear-signing-erc7730-registry/pull/2888

encryption tests should now be ready to be ran again

melanciani:
(re-triggered the CI as it was buggy)

registry/zama/calldata-ConfidentialWrapper.json
```diff
@@ -0,0 +1,201 @@
...
+      "decryptionProof": { "label": "Decryption proof", "format": "raw" },
+      "callbackData": { "label": "Callback data", "format": "raw" }
+    },
+    "formats": {
+      "wrap(address to, uint256 amount)": {
+        "$id": "wrap",
+        "intent": "Shield",
+        "interpolatedIntent": "Shield {amount} underlying units to {to}",
+        "fields": [
+          { "path": "amount", "label": "Amount (underlying units)", "format": "raw" },
```
melanciani, line 49: here, we are forced to use `raw` as it is currently not possible to consider the underlying token's decimals to display this amount. The right approach is most likely to use the `metadata.maps` feature, but this is not supported by the sourcifyeth (and llbartek's lib has a different approach than the EIP). I've thus opened https://github.com/sourcifyeth/clear-signing/pull/67 to suggest adding this feature, and https://github.com/ethereum/ERCs/pull/1966 to clarify the ERC's specs on it  

imo, this does not block this PR and could be fixed in a follow-up PR

melanciani:
hey @manuelwedler, the PR has been updated and is ready for review. I can see that the CI is green for both ts and rust libs: https://github.com/ethereum/clear-signing-erc7730-registry/pull/2595#issuecomment-4893391473 !

kuzdogan:
Thanks for this. I checked the deployments and have two questions before we go further.

**1. Proxies.** All 20 calldata addresses (the 2 ACL and the 18 ConfidentialWrapper deployments) are `ERC1967Proxy` contracts. On mainnet every wrapper proxy points at the same implementation, `0x2ABad2203Eba104b52cf040cCcFA100Df15687F8`, and on Sepolia at `0xAe37b998d453E1FaBE85DD46cf04295ca4A3af04`; both implementations are verified on Sourcify and byte-identical apart from metadata. I don't recall us discussing this case specifically, and how we handle proxies is still unclear to me in general. The spec says descriptors should be bound to the implementation, not the proxy. Since all proxies here share one implementation, would it make sense to bind the wrapper descriptor to the implementation address instead of listing the 18 proxies? Happy to hear how you see it.

**2. EIP-712 verifying contracts.** The three EIP-712 descriptors declare `0x0f6024a97684f7d90ddb0fAAD79cB15F2C888D24` on mainnet and `0x5D8BD78e2ea6bbE41f26dFe9fdaEAa349e077478` on Sepolia. Neither address has code on those chains. Our understanding, from the relayer SDK: the `Decryption` contract lives on Zama's gateway chain, and the SDK builds the EIP-712 domain with the host chainId (1 / 11155111) together with the gateway contract address, so wallets look up exactly the pair the descriptors declare, even though nothing is deployed there on the host chain. Is that accurate? We are writing it down here for future reference, since none of our checks (Sourcify, Etherscan, bytecode) can verify these two entries.

melanciani:
hey again @kuzdogan , great questions indeed: 
1. Could you clarify a bit where you see that specs talk about binding to the implementation ? In the EIP 7730 specs, the upgrade paragraph says that the registry should consider the proxy address : 

```
Upgradeable proxy contracts (such as Transparent Proxies or UUPS proxies as defined in [ERC-1822](https://eips.ethereum.org/EIPS/eip-1822)) use a fixed proxy address that delegates calls to a replaceable implementation contract. For these contracts, the ERC-7730 file SHOULD be bound to the proxy address, since this is the stable address users interact with.
```

imo this makes more sense. Putting the implementation address means anytime we upgrade the proxies, the registry will need to change. Also, the user directly calls/sees the proxy address when signing, not the implementation one, so I'm curious how that would look. But maybe I'm missing something here

2. Yes indeed, I should have made this point more clear. The EIP712 are signed on Ethereum but are actually propagated and verified on the gateway chain (chainIds: mainnet: 261131; testnet: 10901). For this to work, we've updated the way these signatures are verified onchain (https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/contracts/Decryption.sol#L1135). For the record, initially, we were actually making users sign for the gateway chain to avoid this. However, since wallets were still connected to Ethereum when signing the EIP712, they were displaying a big (and very reasonable) warning message as the chainId differed, and we preferred to avoid that ! With that being said: I indeed don't see any proper solution around this convoluted mechanism with the current schema. A suggestion would be to update the spec as such, although I think the `deployments` semantic now becomes misleading + it is very specific to edge cases like us: https://github.com/ethereum/ERCs/pull/1976

hilljessica8787-dot (approved):

kuzdogan:
Thanks for answering those, now it's clearer.

1. I realized I was only referring to the 3rd type of proxies e.g. Safe Proxies where there's a reference implementation and all instances are proxy contracts pointing to the reference. For those it's infeasible to add every single proxy, and here the descriptors need to be bound to the implementation.

For upgradable proxies, the spec indeed tells to point to the proxy, as you point out. I'd still argue this is a bit dangerous since the proxy implementation can move (maliciously or not) and the users might think they are still signing the same transactions. For that we either need to bind the spec to the implementation and resolve proxies on the fly, or implement https://github.com/ethereum/ERCs/pull/1738 such that we have guarantees over the delegated implementation address. 

Even though it's not ideal as is, it seems fine as is and we can figure it out later. 

This discussion also made the proxy issues clearer to me, thank you 

2. All good, and just wanted to double check. In earnest the wording `deployments` is not accurate here. EIP712 messages are completely offchain and there is no deployment necessarily. It's just that usually the verifying contract is on that chainId and address, but as in your case, it can be totally different. It's rather the "domain" as named in EIP712

Anyway it's good as is. Thanks!

kuzdogan (approved):

PR #2614, opened 2026-06-19
feature: add rocket pool contracts to the registry

nickdoherty:
Adds the Rocket Pool protocol contracts to the registry.


manuelwedler:
Hey there :)

manuelwedler:

venugopalanvip:
Reviewed this against the `docs/REVIEWING.md` checklist. Not a maintainer — just a contributor trying to help the queue move, since #2871 names review capacity rather than submissions as the constraint. I closed my own #2961 in favour of this PR: it described `rETH` at the same address, three months later and with less coverage.

**Checklist**

1. **CI** — green across all 18 descriptors, both runners.
2. **ABI fetched** — no "Could not fetch ABI" warnings, so display fields were validated against the contracts.
3. **One entity per PR** — every file is under `registry/rocketpool/`.
4. **Intents and hidden fields** — I read all 18. Intents name actions in user terms throughout (`Redeem rETH for ETH`, `Stake RPL`, `Claim rewards`), no bare function names, and I found no value-bearing parameter marked `visible: "never"`. All intents are within 30 characters and all labels within 20.

Nice work — this is more consistent than most of what is in the registry today.

**One finding, non-blocking**

`increaseAllowance` has no `threshold` on its amount field, while `approve` in the same file does:

```
registry/rocketpool/calldata-rocketTokenRETH.json
  approve            "params": { "tokenPath": "@.to", "threshold": "0x8000…", "message": "Unlimited" }
  increaseAllowance  "params": { "tokenPath": "@.to" }
```

Same asymmetry in `calldata-rocketTokenRPL.json`. So an unlimited `approve` reads as `Unlimited rETH`, but an unlimited `increaseAllowance` renders as a 78-digit integer.

This is not hypothetical. [`0xa16ed231ae9d28e0acd4030f0a38da006a9203805f4569a4e68a348a7618f173`](https://etherscan.io/tx/0xa16ed231ae9d28e0acd4030f0a38da006a9203805f4569a4e68a348a7618f173) is a mainnet `increaseAllowance` of `type(uint256).max` on rETH. I used it as a test fixture and the runners rendered:

```
Increase by 115792089237316195423570985008687907853269984665640564039457.584007913129639935 rETH
```

Adding the same `threshold` and `message` that `approve` already uses turns that into `Unlimited rETH`. Cyfrin raised exactly this against Swell in #2815, and the registry's convention is the 2^255 value you have already used for `approve`, so it is a two-line change in each of the two token descriptors.

`decreaseAllowance` has the same gap but I would leave it alone — an unlimited *decrease* clears an allowance rather than granting one, so labelling it "Unlimited" would describe a safe action in the language used for a dangerous one.

Happy to open a PR against your branch with the change if that is easier than doing it yourselves.

PR #2624, opened 2026-07-03
Update V2 schema with json-schema 2020-12

melanciani:
Following fix introduced in https://github.com/ethereum/ERCs/pull/1822 (once merged

note:

- I've made Claude run a check on all current schemas in the registry repo and only a single schema fails, but because of a pre-existing unrelated issue : `registry/paraswap/calldata-AugustusSwapper-v5.json`
- this can be checked by running `check-jsonschema --schemafile specs/erc7730-v2.schema.json registry/paraswap/calldata-AugustusSwapper-v5.json` in main currently

### DIFF
--- a/specs/erc7730-v2.schema.json
+++ b/specs/erc7730-v2.schema.json
@@ -348,15 +348,7 @@
                     "description": "The list includes formatting info for each field of a structure. For contract bindings, entries are keyed by the full function signature with parameter names; for EIP712 bindings, entries are keyed by the string returned by EIP 712 encodeType on the primary type.",
                     "type": "object",
                     "propertyNames": {
-                        "anyOf": [
-                            {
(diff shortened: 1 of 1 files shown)


melanciani:
about the paraswap pre-existing issue, I think it should be fixed in another PR. Claude suggests this : 

it's **tuple-nesting depth**, nothing to do with your dialect change:

- The regex's tuple matcher `(?:[^()]|\([^()]*\))*` only allows **one** level of nested parentheses.
- `multiSwap` nests **3** levels deep, `megaSwap` **4** (path → adapters → route, plus mega's extra `fromAmountPercent` wrapper). My test shows both start matching only at depth 3/4 respectively; every other paraswap signature matches at depth 1.

So the schema regex is simply too shallow for paraswap's deeply nested structs. Here are the fix options, best first:

## Option 1 — Bump the nesting depth in the regex (recommended, keeps strictness)
Nest the alternation to ~4 levels. Replace the tuple sub-pattern
`(?:tuple)?\s*\((?:[^()]|\([^()]*\))*\)`
with
`(?:tuple)?\s*\((?:[^()]|\((?:[^()]|\((?:[^()]|\([^()]*\))*\))*\))*\)`
(applied to **both** occurrences on line 353). I verified a depth-4 pattern matches all 16 paraswap signatures. Downside: ugly, and it's still bounded — a hypothetical 5-deep struct would fail. But real ABIs almost never exceed this.
Caveat: JSON Schema 2020-12 mandates ECMA-262 regex, which has **no recursion**, so *unbounded* depth is impossible by spec — bounded nesting is the strictest achievable.

## Option 2 — Relax to coarse shape + lean on the linter
Change the pattern to just `^[A-Za-z_][A-Za-z0-9_:]*\(.*\)$` (name + parens). The authoritative validator is the `python-erc7730` linter, which actually parses the ABI; the schema regex is a redundant gate. This is future-proof against any nesting but drops schema-level structural strictness.

kuzdogan:
Let's move forward with option 2

melanciani:
PR to fix the nested tupple issue: https://github.com/ethereum/ERCs/pull/1868

kuzdogan (approved):

kuzdogan:
Thank you!

PR #2634, opened 2026-07-16
Keep index files in sync

### LINKED ISSUE #2619: CI check for index updates
A few newer descriptors are missing in the two index files because the indexes were generated before some descriptors were added:
- https://github.com/ethereum/clear-signing-erc7730-registry/blob/master/index.calldata.json
- https://github.com/ethereum/clear-signing-erc7730-registry/blob/master/index.eip712.json

We should make PRs fail via the CI if they don't add an entry to the index files for any new descriptors.

Also add some documentation in the README about the index files.

manuelwedler:
Fixes #2619 

Regenerates the drifted index files and adds CI to keep them in sync.

- **Regenerate the index files** — `index.calldata.json` was missing 142 entries; `index.eip712.json` was missing 2 and still pointed at 6 deleted descriptors.
- **Add a generator** — `npm run generate-index` derives both files from the descriptors, resolving `includes` first.
- **Add a weekly workflow** — regenerates the index and opens a PR only when it changed, so contributors never have to update it themselves.
- **Document** the index files at the end of the README, trimmed to what consumers need.
- **Align schedules** — all recurring workflows now run weekly on Mondays at 03:00 UTC.


kuzdogan:
As is, when someone opens a PR to add a descriptor, they also need to generate the index within the PR. From my understanding this needs to be done manually. I think this is bad UX. The index generation run automatically after the PR is merged by itself or it should run automatically on the PR adding the descriptor.

manuelwedler:
**Writing directly to `master`:**
- `master` is protected (PRs required), so a CI push is rejected unless we grant the Actions bot a branch-protection bypass — which weakens the protection.
- Auto-commits bypass review, and pushing from CI can retrigger workflows (loop risk).
- A bad or non-deterministic regeneration would land on `master` unreviewed, with no PR to catch it first.

**Writing to fork PRs:**
- The `GITHUB_TOKEN` is scoped to the base repo and cannot push to a fork's branch, so we'd need a dedicated bot PAT/App token (the same credential we lost in the handover).
- It also requires the contributor's "Allow edits by maintainers" (per-PR, opt-out, and unavailable for org-owned forks), and running it safely means the `pull_request_target` untrusted-code handling.

**Alternative:** Instead, we could add a scheduled workflow that regenerates the index and opens a PR — exactly like the existing `sync-specs` and `format` jobs. It runs on `master` (trusted, no fork-token or untrusted-code issues), keeps `master` authoritative, and leaves a normal PR for a maintainer to merge. The PR-time index check can then be relaxed to a warning, so contributors never have to regenerate the index by hand.

~~**Caveat:** One benefit of updating the index at PR time (by the submitter) is that calldata formatters may rely on the index, so it might need to be up to date for the clear-signing tests to run correctly within the PR. Moving index generation off the PR would lose that.~~ This assumption was wrong. The test runners actually build the index themselves by walking through the registry repo at the current commit.

@kuzdogan what do you think?

manuelwedler:
Updated to use a scheduled workflow for index generation

kuzdogan:
IMO the generator should run more often, if not on each push to master

kuzdogan:
A couple of notes on the workflow/tooling side:

1. **`--check` isn't wired into CI** — nothing runs `generate-index --check`, so index drift on a contributor PR is only caught by the weekly `sync-indexes` job. Fine if it's meant as a local-only helper; a quick comment saying so would clarify intent.
2. **`format.yml` went daily → weekly** — `0 7 * * *` → `0 3 * * 1` is a 7× reduction in formatting frequency, not just a time shift. Just confirming the reduced cadence is intended, not incidental to the schedule alignment.

manuelwedler:
> A couple of notes on the workflow/tooling side:
> 
> 1. **`--check` isn't wired into CI** — nothing runs `generate-index --check`, so index drift on a contributor PR is only caught by the weekly `sync-indexes` job. Fine if it's meant as a local-only helper; a quick comment saying so would clarify intent.
> 2. **`format.yml` went daily → weekly** — `0 7 * * *` → `0 3 * * 1` is a 7× reduction in formatting frequency, not just a time shift. Just confirming the reduced cadence is intended, not incidental to the schedule alignment.
> 
> Posted with Claude Code

1. It's still there from the CI check I removed now, but I think it's good to keep as a helper.
2. I wanted to align the schedule of the PRs, and this shift was intended. I think it creates an overhead if get a formatting PR in every day.

manuelwedler:
> IMO the generator should run more often, if not on each push to master

Updated to regenerate the indexes on every push to master

kuzdogan:
Reviewed the logic — it checks out. I ran the generator against the branch and it reproduces the committed index files exactly (idempotent), `includes`-based descriptors resolve and get indexed correctly, and no eip712 or calldata descriptors are silently skipped. Two minor, non-blocking notes:

### 1. What should happen on a calldata collision?

In `tools/scripts/generate-index.js`:

```js
if (!index.calldata[key]) index.calldata[key] = descriptorPath;
```

When two descriptors declare a deployment at the same `(chainId, address)`, the current behavior is "first sorted path wins, the rest are dropped silently." There are **no such collisions in the registry today**, so this isn't an active bug — but it's an undefined case worth deciding on deliberately. IMO we should throw in such cases.

Also about this, we talked about having multiple descriptors for different schema versions. We should keep this in mind.

### 2. `ercs/**` doesn't feed the index

In `.github/workflows/sync-indexes.yml`:

```yaml
    paths:
      - "registry/**"
      - "ercs/**"
```

The generator only walks `registry/` (`findDescriptors(REGISTRY_DIR)`), and no descriptor's `includes` resolves into `ercs/`, so a push touching only `ercs/` triggers this workflow, regenerates identical files, and no-ops (wasted runner, never a PR). Suggest dropping `ercs/**` — or, if the intent is that spec/schema changes under `ercs/` *should* affect index output, that isn't wired up today and a comment explaining the intent would help.

kuzdogan (approved):

kuzdogan (dismissed):
Approving with minor comments

manuelwedler:
### Regarding 1.

Good catch! I agree that we should throw when we encounter such a case. I added this change.

We should also make the CI fail on pull requests, if a new descriptor would create a collision on an index entry. But I will implement this in another PR.

### Regarding 2.

Good catch to flag it, but `ercs/**` should stay — it *does* feed the index:

- **180 descriptors `include` files from `ercs/`**: 74 permit descriptors → `ercs/eip712-erc2612-permit.json`, 106 vault descriptors → `ercs/calldata-erc4626-vaults.json`. The generator resolves includes before indexing, so that content flows in.
- **Verified empirically**: changing one param name in `ercs/eip712-erc2612-permit.json` regenerates a *different* `index.eip712.json` (the `encodeTypeHashes` flip, since they're `keccak256(format key)` and that key comes from the `ercs/` file via `includes`).

You're right that the generator only *walks* `registry/` — but registry descriptors reach into `ercs/` via `includes`, so an `ercs/` edit can legitimately change the index. Dropping `ercs/**` would leave the published index stale after an `ercs/` change.

(Minor nuance: the calldata index is deployment-keyed, so `ercs/` format changes only affect the eip712 index — but that's enough to justify the trigger.)

manuelwedler:
@kuzdogan this needs a new approval due to the added commit

kuzdogan (approved):

kuzdogan:
@manuelwedler ah nice, I guess it's a different repo setting that voids approvals on added commits. We should maybe have this on our repos too

Approved

Issue #2628, opened 2026-07-07
Fix broken CI jobs after EF handover


manuelwedler:
A couple of CI jobs broke due to missing keys/secrets and rights to external repos (https://github.com/ethereum/clear-signing-erc7730-registry/issues/2459). The problems I discovered so far:

- Descriptor validation jobs fail due to missing Etherscan keys (for example https://github.com/ethereum/clear-signing-erc7730-registry/actions/runs/28809724485/job/85434236147)
- Descriptors fail schema validation on master since some descriptors seem incompatible with the schema (https://github.com/ethereum/clear-signing-erc7730-registry/actions/runs/28516583896/job/84529735146)
- "trigger updates" job fails on master due to a missing token (https://github.com/ethereum/clear-signing-erc7730-registry/actions/runs/28516583896/job/84529735146). I'm not sure for what this job even is needed.
- If ledger tests cannot be run due to permissions, disable them (https://github.com/ethereum/clear-signing-erc7730-registry/issues/2459). Also they seem to be broken due to another recent change (https://github.com/ethereum/clear-signing-erc7730-registry/actions/runs/28795144962/job/85383568221). We might also be able to remove the label based gate on the TypeScript and Rust tests when ledger tests are disabled completely.

manuelwedler:
Done

PR #2674, opened 2026-07-22
ci: fix the master workflow

manuelwedler:
See #2628 

Fixes the `master` workflow, broken after the EF handover.

- Remove the obsolete `trigger-updates` job — it dispatched to `LedgerHQ/*` repos and can't authenticate anymore (was already marked for removal once ownership transferred).
- Scope the descriptor lint to real descriptors, excluding the `tests/`/`testsv2/` fixtures — linting the whole `registry/` tree was failing with ~2.6k errors from validating test files as descriptors.
- Wire `ETHERSCAN_API_KEY` into the lint step so ABI validation works (the key is already added to the CI secrets).
- Fix the `erc7730 lint` commands in the READMEs to exclude test fixtures.

## Note: lint will still report errors on master

The lint will flag `Invalid display field` errors on two descriptors: `aave/calldata-WrappedTokenGatewayV3` (`#.pool`) and `1inch/calldata-AggregationRouterV3` (`#.pools`).

These descriptors declare their function signatures **with named parameters** — e.g. `depositETH(address pool, address onBehalfOf, uint16 referralCode)` — and display fields reference those names (`#.pool`). But the linter validates the paths against the **on-chain ABI fetched from Etherscan**, where the same parameters are **unnamed** (`depositETH(address, address, uint16)`). Since the name isn't in the fetched ABI, the path can't be resolved and the check errors — even though the descriptor names it correctly.

So the descriptors are fine; the linter is wrong to require a name the on-chain ABI never provides. We'll fix this on the linter side (`python-erc7730`).


README.md
```diff
@@ -51,11 +51,11 @@ The `erc7730` Python package is available for validating and formatting ERC-7730
...
 
-# Validate all descriptors
-erc7730 lint registry/**/eip712-*.json registry/**/calldata-*.json
-
 # Validate a specific file
-erc7730 lint registry/entity/calldata-Contract.json
+erc7730 lint registry/uniswap/calldata-UniswapV3Router02.json
+
+# Validate all descriptors (exclude the *.tests.json fixtures, which are not descriptors)
+erc7730 lint $(find registry -type f \( -name 'calldata-*.json' -o -name 'eip712-*.json' \) -not -name '*.tests.json')
```
kuzdogan: Minor: this excludes fixtures by filename (`*.tests.json`), while `master.yml` excludes them by directory (`tests/`, `testsv2/`). Same result today, but worth matching them so the documented command and CI stay in sync.

.github/workflows/master.yml
kuzdogan: Do we need this branch? Or is it for live CI testing?

manuelwedler: seem like a legacy branch. I'll remove it

kuzdogan:
Looks like `setup.sh` is not fixed?

Edit: My bad, my review was left pending...

kuzdogan (changes requested):

kuzdogan:
Nice fix — the reasoning is clear and the change is correct. Two follow-ups plus a small nit:

**1. `tools/linter/setup.sh` still prints the two things this PR fixes.**
The script echoes usage instructions after setup, and those lines weren't updated:
- `setup.sh:92` and `:95` print `erc7730 lint registry/` — the exact broken command that sweeps in the test fixtures and fails.
- `setup.sh:98` prints `erc7730 lint registry/uniswap/calldata-permit2.json`, which is a non-existent file (this PR already replaced it with `calldata-UniswapV3Router02.json` in the READMEs).

So a contributor who follows the setup script's own printed output hits exactly the failure this PR aims to remove. Could we update those lines here too? (`tools/scripts/generate-7730.README.md:255` uses `erc7730 lint registry/myprotocol/` — lower priority, but same theme.)

**2. Three different fixture-exclusion patterns now coexist.**
- `pull_request.yml` excludes `tests/**`, `testsv2/**`, *and* `*.tests.json`
- `master.yml` (this PR) excludes by path: `tests/`, `testsv2/`
- `README.md` (this PR) excludes by name: `*.tests.json`

All three give the same result today, so nothing's broken — but the workflow and docs now teach different commands, and neither matches the existing `pull_request.yml` convention. Aligning them to one set (ideally `pull_request.yml`'s) would be more robust if the conventions ever drift.

**3. Nit (optional): unquoted `$FILES` word-splitting in `master.yml`.**
`erc7730 lint $FILES --gha` relies on word-splitting, which is fine today since no registry path contains spaces. `find … -print0 | xargs -0 erc7730 lint --gha` would avoid both that assumption and the (currently impossible) empty-list case. Not necessary, just noting.

kuzdogan (approved):

PR #2765, opened 2026-07-29
Add Cyfrin attestations for 181 reviewed descriptors

0kage-eth:
Adds 181 EAS offchain attestations (ERC-8176 schema
0xe023eef113c1670774801c34b377fdf612dd8a4d2fa92fe382e15bd91fafb5c2) from
Cyfrin's clear-signing review, signed by Patrick Collins
(0x3846c3A30E62075Fa916216b35EF04B8F53931f6 — see the accompanying auditor profile PR #2764).

Each descriptor was reviewed per the auditors/README.md 5-step process
(project check, Sourcify contract verification, descriptor accuracy,
intent-mutability, translate validation). For every attestation:
- message.data equals the descriptor's ERC-8176 hash (clearsig dh <descriptor>),
- the signature verifies via the EAS SDK verifyOffchainAttestationSignature,
- the recovered signer is the address above.

Additions only: 181 new registry/<project>/sigs/*.eip155-1-0x3846...931f6.json
files, no descriptor changes.

Coverage (181): morpho 104, kiln 29, flare 4, threshold 4, ondo-finance 3,
lido 3, 1inch 3, safe 3, walletconnect 2, swissborg 2, sei 2, p2p 2,
corestake 2, circle 2, celo 2, and one each for hyperliquid, yieldxyz, weth,
tether, starkgate, poap, paraswap, ledgerquest, layerswap, igra, figment,
ethena, consensus-specs, avvatar-labs-alia.

Update 2026-07-31: added 6 attestations for descriptors that passed review
after the initial batch (hyperliquid eip712-withdraw, p2p EigenPodManager and
P2pMessageSender, safe eip712-Safe-Multisig, threshold
L1BitcoinDepositor-address and L2BitcoinDepositor).

PR #2293, opened 2026-01-27
add more Starkgate bridge contracts

adrienlacombe:
## Summary

Adds ERC-7730 v2 clear-signing metadata for Starkgate L1 bridges on Ethereum mainnet, complementing the existing STRK descriptor.

- Owner: `StarkWare` (MakerDAO `L1DAIBridge` is a separate descriptor)
- No deprecated Etherscan ABI URLs
- `deploymentDate` set from each contract's creation transaction
- `interpolatedIntent`: `Bridge {amount} to {l2Recipient}`

## Bridges covered

AAVE, DAI (current StarkGate + MakerDAO v0), ENA, ETH, EURC, FRAX, FXS, GGMT, LBTC, LINK, LUSD, MultiToken, R, SolvBTC, UNI, USDC, USDT, USR, WBTC, ZRO, brBTC, enzoBTC, mRe7BTC, pumpBTC, rETH, sfrxETH, tBTC, uniBTC, wstETH

Addresses match [starknet-io/starknet-addresses](https://github.com/starknet-io/starknet-addresses) `bridged_tokens/mainnet.json`.

`calldata-StarkGate-STRK.json` is updated in place (attested; another attestation round is planned).

## Deposit formats

- **3-arg** `deposit(address token, uint256 amount, uint256 l2Recipient)` on every StarkWare proxy
- **Legacy 2-arg** `deposit(uint256 amount, uint256 l2Recipient)` on upgraded StarkGate 1.0 bridges (ETH, FRAX, FXS, LUSD, R, STRK, UNI, USDC, USDT, rETH, sfrxETH, wstETH) and on MakerDAO `L1DAIBridge`
- **`depositWithMessage(...)`** on every StarkWare proxy

## Tests

`testsv2/` fixtures for every new descriptor. Most use a real mainnet `deposit` transaction. **USR**, **pumpBTC**, **EURC**, and **DAI-v0** use unsigned clones (no on-chain 3-arg/2-arg user deposit of the named token was available).

## Test plan

- [x] `erc7730 lint` / JSON schema validation on descriptors and `testsv2` files
- [x] Sourcify + Rust runners passed on `9612702`; re-run on this commit
- [ ] `validate attested descriptors` will fail on STRK (existing attestation; updating in place per review)

PR #76, opened 2024-11-06
add StarkGate STRK bridge contract

adrienlacombe:

PR #2869, opened 2026-08-07
fix(Starkgate): Add another deployment

PrisionMike:
https://etherscan.io/address/0xae0ee0a63a2ce6baeeffe56e7714fb4efe48d419

### DIFF
--- a/registry/starkgate/calldata-StarkGate-STRK.json
+++ b/registry/starkgate/calldata-StarkGate-STRK.json
@@ -2,7 +2,12 @@
   "$schema": "../../specs/erc7730-v2.schema.json",
   "context": {
     "$id": "L1StarkGateSTRKbridge",
-    "contract": { "deployments": [{ "chainId": 1, "address": "0xcE5485Cfb26914C5dcE00B9BAF0580364daFC7a4" }] }
+    "contract": {
+      "deployments": [
+          { "chainId": 1, "address": "0xcE5485Cfb26914C5dcE00B9BAF0580364daFC7a4" },
+          { "chainId": 1, "address": "0xae0Ee0A63A2cE6BaeEFFE56e7714FB4EFE48D419" }
+        ]
+      }
   },
   "metadata": {
     "owner": "StarkWare",


kuzdogan:
Thanks for this — but I don't think this address belongs in this descriptor. `0xae0Ee0A6…E48D419` isn't another deployment of the STRK bridge; it's StarkGate's **ETH** bridge.

Both addresses are proxies, but they forward to different implementations:

| | `0xcE5485Cf…daFC7a4` (existing) | `0xae0Ee0A6…E48D419` (this PR) |
|---|---|---|
| implementation | `0x6ad74d4b…fdc85` | `0x95ff25a5…7fd95` |
| verified as | `StarknetERC20Bridge` | `StarknetEthBridge` |
| `token` arg accepted by `deposit` | STRK `0xca14007e…42766` | placeholder `0x…455448` only |
| passing STRK | works | reverts `TOKEN_NOT_SERVICED` |
| `msg.value` means | the fee | bridged ETH **+** fee |

They service strictly opposite tokens (`getStatus`: STRK is active on one and unknown on the other, and vice versa). Both do declare `deposit(address,uint256,uint256)` — so the descriptor applies cleanly and nothing fails validation. It just renders wrong.

### What the user would see

Simulated a real, succeeding ETH deposit on `0xae0Ee0A6…`: bridging 1 ETH with a 0.0001 ETH fee, so `msg.value` = 1.0001 ETH.

| descriptor shows | reality |
|---|---|
| Amount to deposit: `1000000000000000000` (no ticker) | 1 ETH |
| Recipient: `627…` | ✓ correct |
| **Deposit fee: 1.0001 ETH** | **0.0001 ETH** |

Two causes:

1. `"format": "tokenAmount"` with `"tokenPath": "token"` makes the wallet read ticker and decimals from the `token` argument. On the ETH bridge that's always `0x…455448` — `address(0x455448)`, the ASCII for "ETH", declared in `StarkgateConstants.sol`. Nothing is deployed there, so there's nothing to read.

2. `@.value` is labelled `"Deposit fee"`. That's right for the ERC-20 bridge, whose `acceptDeposit` returns `msg.value` as the fee outright. `StarknetEthBridge` overrides it:

   ```solidity
   require(msg.value >= amount, "INSUFFICIENT_VALUE");
   uint256 fee = msg.value - amount;
   ```

   So the fee line reports the bridged principal plus the fee — here, ~10,000× the real cost, and close enough to the deposit amount that it reads as "the bridge is taking your whole deposit."

### This also invalidates the Cyfrin attestation

This descriptor is one of the 181 signed in #2765 (Cyfrin attestations for reviewed descriptors), at `registry/starkgate/sigs/calldata-StarkGate-STRK.eip155-1-0x3846c3A3…931f6.json`.

The ERC-8176 hash is `keccak256` over the canonicalized descriptor JSON, so it covers `context.contract.deployments`:

- attested `message.data`: `0x55a2695788101c32…be704cf2`
- master today: `0x55a2695788101c32…be704cf2` ✅
- this branch: `0xb7059e529ff71cb5…93a8044e` ❌

Merging this leaves a signature that no longer covers the file it sits next to.

### Suggested path

A separate `calldata-StarkGate-ETH.json` for the ETH bridge, with a native-ETH amount format instead of a `tokenPath` lookup, a fee field derived from `value - amount` rather than `@.value`, and intent text naming ETH. Worth also covering `deposit(uint256,uint256)` (`0xe2bbb158`), the legacy two-arg entry point both bridges expose and neither descriptor describes.

hilljessica8787-dot (approved):

PrisionMike:
Closing in favour of #100 #76 #2293  and other Starkgate PRs. Can revisit after they are merged.

PR #2870, opened 2026-08-10
Clip threshold values

PrisionMike:
Spec recommends threshold to be activated at [ABOVE](https://eips.ethereum.org/EIPS/eip-7730#:~:text=integer%20value%2C%20above%20which%20value%20is%20displayed%20as%20a%20special%20message%2E%20Optional) the threshold. By definition it should never be `UINT256_MAX`

Would recommend `0x8000000000000000000000000000000000000000000000000000000000000000` in general but that could beget a confirmation from the providers.


kuzdogan (approved):
Approving. Also does not affect attestations in #2765

Issue #2747, opened 2026-07-29
Cyfrin Audit: [Permit - eip712-permit-avalanche_c_chain-yetiswap]


0kage-eth:
**Descriptor:** [registry/permit/eip712-permit-avalanche_c_chain-yetiswap.json](https://github.com/ethereum/clear-signing-erc7730-registry/blob/a2b33ffed92aab99cb9da2704610469a2588b520/registry/permit/eip712-permit-avalanche_c_chain-yetiswap.json)

**ERC-8176 descriptor hash:** `0x0b734fdabf59cc7f1352210fa629108acabed5f4d26c6113985921ac62519941`

**1: `value` has no `threshold`, so an unlimited approval renders as a 78-digit number**
**Severity:** Medium

**Description:** Unlimited permits are the highest-risk case; without a `threshold`/`message` the user gets no "Unlimited" cue and cannot recognize an unbounded allowance.

**Recommended mitigation:** Add to the `value` field's `params`:

```diff
// ercs/eip712-erc2612-permit.json:10-16
+"threshold": "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
+"message": "Unlimited"
```

**2: YetiSwap does not expose a standard ERC-2612 `DOMAIN_SEPARATOR()`, so the descriptor cannot be bound to the live permit domain**
**Severity:** Medium

**Description:** The descriptor declares a standard ERC-2612 `Permit(...)` domain, but the deployed YTS token does not expose the standard `DOMAIN_SEPARATOR()` getter used to verify that domain against live accepted signatures. This is the same fail-closed binding class carried by no-standard-permit siblings in the permit family: the descriptor may be structurally valid, but clear-signing cannot be trusted to apply to an accepted live permit without a verifiable on-chain domain.

**Recommended mitigation:** Replace this descriptor with one that matches YTS's actual permit implementation and domain-verification surface, or withdraw the descriptor if YTS does not implement a verifiable ERC-2612-compatible permit.

```diff
// registry/permit/eip712-permit-avalanche_c_chain-yetiswap.json
- registry/permit/eip712-permit-avalanche_c_chain-yetiswap.json
+ remove this descriptor, or replace it with a descriptor for the token's actual signed message type
```

kuzdogan:
## Finding 1 re-checked — YetiSwap does implement ERC-2612

[...]

PR #2890, opened 2026-08-14
fix(permit): remove non-existent version field from YetiSwap test domain type

kuzdogan:
Follow-up to #2747 (YetiSwap permit descriptor).

`registry/permit/testsv2/eip712-permit-avalanche_c_chain-yetiswap.tests.json` declared a four-field EIP-712 domain type but supplied three values. YetiSwap has no `version` field — its `DOMAIN_TYPEHASH` is `EIP712Domain(string name,uint256 chainId,address verifyingContract)` (verified on Sourcify: chain 43114, `0x488f73cddda1de3664775ffd91623637383d6404`, full match).

The two types give different separators:

| Built from | Domain separator |
|---|---|
| the contract's real three-field type | `0x8d5fecbb9fafdb8768a4e906ac8751d1c7ded665e93286e8d5f104b5c3b7f5cd` |
| the test vector's four-field type | `0x24134fc0ec87b29dc85d76bab6aaada501d1628435ee678932dfdbbbf6bb5de4` |

So the vector described a message the contract would reject with `YTS::permit: unauthorized`.

CI never caught this because the runners assert rendered output only, and the domain match reads the `domain` values rather than the `types` list. It is a fidelity defect in test data, not a test failure — which is why it needs a deliberate fix.

One line removed. The descriptor itself is correct and untouched.

**The v1 fixture is left alone deliberately.** `registry/permit/tests/eip712-permit-avalanche_c_chain-yetiswap.tests.json` both declares and supplies `version`, so it is internally consistent; removing the type entry would create the mirror-image mismatch. Its domain is placeholder data (`"name": "Example"`) that does not describe the real contract — a broader problem shared with other v1 fixtures, and worth a separate change.

### DIFF
--- a/registry/permit/testsv2/eip712-permit-avalanche_c_chain-yetiswap.tests.json
+++ b/registry/permit/testsv2/eip712-permit-avalanche_c_chain-yetiswap.tests.json
@@ -8,7 +8,6 @@
         "types": {
           "EIP712Domain": [
             { "name": "name", "type": "string" },
-            { "name": "version", "type": "string" },
             { "name": "chainId", "type": "uint256" },
             { "name": "verifyingContract", "type": "address" }
           ],


fbwoolf (approved):
Approving. One line removed, and the reasoning behind it is sound.

I checked the fixture on master: the `EIP712Domain` type list declares four members while the `domain` object supplies three, `name`, `chainId` and `verifyingContract`. So the vector described a message built from a domain type the token does not use, and the separator it implies is not the one the contract enforces. Removing the `version` entry makes the declared type match both the supplied values and the contract's `DOMAIN_TYPEHASH`.

**Leaving the v1 fixture alone is the right call**, and worth stating plainly so it does not read as an oversight. `registry/permit/tests/eip712-permit-avalanche_c_chain-yetiswap.tests.json` both declares and supplies `version`, so it is internally consistent as it stands. Removing the type entry there would create the mirror image of the bug being fixed here. Its domain is placeholder data (`"name": "Example"`) that describes no real contract, which is a wider problem across the v1 fixtures and belongs in its own change.

**Verified locally** against `@ethereum-sourcify/clear-signing-test-runner` at the pin this repository uses (`dae3cda`, v0.2.2): 1 case, 1 pass.

The observation in the description is the durable part: CI did not catch this because the runners assert rendered output, and the domain match reads the `domain` values rather than the `types` list. That is the same blind spot #2896 describes, and it is the gap the domain-separator check proposed there would close. This PR fixes one instance of it by hand; the check would find the rest.

No audit issue closes here. #2747 was already closed on the verification recorded on its thread, and this is a fidelity fix to the fixture that verification produced.

PR #2891, opened 2026-08-14
fix(uniswap): show max Permit2 amounts as "Unlimited"

### LINKED ISSUE #2677: Cyfrin Audit: [Uniswap - eip712-uniswap-permit2]
**Descriptor:** [registry/uniswap/eip712-uniswap-permit2.json](https://github.com/ethereum/clear-signing-erc7730-registry/blob/08b9d5177d08109c6dbacde2c00220ef6129ba64/registry/uniswap/eip712-uniswap-permit2.json)

**ERC-8176 descriptor hash:** `0x79d6ea32b32fbfbfc045cdc2c698b65733062ae79208ae4f64956a1157a817af`

**1: Max Permit2 amounts render as giant numbers instead of a clear max-value cue**

**Severity:** Medium

**Description:** on the allowance forms, the signer grants an unbounded persistent allowance but sees a giant integer instead of "Unlimited". On `PermitTransferFrom`, the field is not a persistent allowance, but a max `uint256` transfer cap still needs an explicit max-value cue so the one-shot cap is understandable.

**Recommended mitigation:** The threshold is sized to `type(uint160).max` (40 hex chars), not `type(uint256).max`, because the field is `uint160`. Apply the same change to `details.[].token` in PermitBatch.

For `PermitTransferFrom`, use the `uint256` max threshold and one-shot wording:

```diff
// registry/uniswap/eip712-uniswap-permit2.json:6
- "PermitSingle(PermitDetails details,address spender,uint256 sigDeadline)PermitDetails(address token,uint160 amount,uint48 expiration,uint48 nonce)": {

+"params": {
+"tokenPath": "details.token",
+"threshold": "0xffffffffffffffffffffffffffffffffffffffff",
+"message": "Unlimited"
+}
```

**2: Two production deployments are unverified on Sourcify**

**Severity:** Medium

**Description:** Sourcify reports no verified source at the Permit2 address on production chain 146 (Sonic) and production chain 81457 (Blast). An attestation vouches for every listed production deployment, so those deployments block signing until they are verified on Sourcify.

**Recommended mitigation:** Verify the Permit2 deployment on chains 146 and 81457, or remove those production-chain deployments from the descriptor if they are not in scope.

kuzdogan:
## The problem

Permit2 is a Uniswap contract. It lets a user give a token allowance with a signature. The user does not send a transaction. Permit2 has three messages that a user can sign. Each message contains an amount.

If the amount is the maximum value, the allowance has no limit. Today the wallet shows that maximum as a very long number. The user sees 46 digits. The user cannot tell that the allowance has no limit. This is audit finding #2677.

ERC-7730 has two parameters for this problem. The `threshold` parameter holds a value. The `message` parameter holds a text. If the amount reaches the threshold, the wallet shows the text in place of the number.

## Why this change uses two different threshold values

The three fields do not have the same type. Therefore they do not get the same threshold.

**The two allowance fields.** `PermitSingle.details.amount` and `PermitBatch.details.[].amount` are `uint160`. Their threshold is `type(uint160).max`. That value is 40 hex characters, not 64.

This value is also the true sentinel of Permit2. The function `AllowanceTransfer._transfer` keeps the allowance at its full value only when the allowance is equal to `type(uint160).max`:

```solidity
uint256 maxAmount = allowed.amount;
if (maxAmount != type(uint160).max) {
    ...
    allowed.amount = uint160(maxAmount) - amount;
}
```

Each transfer decreases a smaller allowance. Therefore only the maximum value gives an allowance with no limit.

**The transfer field.** `PermitTransferFrom.permitted.amount` is `uint256`. This field is a limit for one transfer only. It is not an allowance that stays after the transfer. Its threshold is `2^255`. The file `ercs/eip712-erc2612-permit.json` uses the same value. The text is "Any amount" and not "Unlimited", because "Unlimited" can make the reader believe that the allowance stays.

## The threshold also fires on an equal value

A reader can ask if an exact `type(uint160).max` threshold ever fires. The specification answers this question. In `specs/erc-7730.md`, the `tokenAmount` example gives the field value `0xFFFFFFFF` and the threshold `0xFFFFFFFF`. The result is `Unlimited DAI`. Therefore the comparison includes the equal value.

## Tests

This change adds three test cases:

- A `PermitSingle` message with an amount that has no limit.
- A `PermitBatch` message with one unlimited entry and one normal entry. This case proves that each entry gets its own result.
- A `PermitTransferFrom` message with the maximum `uint256` amount.

Before this change, `PermitTransferFrom` had no test in `testsv2`.

I ran the tests with the Sourcify test runner that CI uses (`clear-signing-test-runner` at `dae3cda`, v0.2.2). The result is 5 pass, 0 fail.

I also ran a negative control. I kept the new tests, and I removed the change from the descriptor. The three new cases then fail. They show the problem that the audit reports:

```
1461501637330902918203684832716283019655932.542975 USDC
Transfer 115792089237316195423570985008687907853269984665640564039457584007913129.639935 USDC
```

## About the issue

Issue #2677 contains a second finding. That finding says that Permit2 has no verified source on Sonic (chain 146) and on Blast (chain 81457). Sourcify verified both contracts on 2026-07-30. I confirmed this against the live API. Therefore this pull request closes the complete issue.

Closes #2677

### DIFF
--- a/registry/uniswap/eip712-uniswap-permit2.json
+++ b/registry/uniswap/eip712-uniswap-permit2.json
@@ -12,7 +12,11 @@
             "path": "details.amount",
             "label": "Amount allowance",
             "format": "tokenAmount",
-            "params": { "tokenPath": "details.token" },
+            "params": {
+              "tokenPath": "details.token",
+              "threshold": "0xffffffffffffffffffffffffffffffffffffffff",
+              "message": "Unlimited"
+            },
             "visible": "always"
           },
           { "path": "details.expiration", "label": "Approval expires", "format": "date", "params": { "encoding": "timestamp" } },
@@ -28,7 +32,11 @@
             "path": "details.[].amount",
             "label": "Amount allowance",
             "format": "tokenAmount",
-            "params": { "tokenPath": "details.[].token" }
+            "params": {
+              "tokenPath": "details.[].token",
+              "threshold": "0xffffffffffffffffffffffffffffffffffffffff",
+              "message": "Unlimited"
+            }
           },
           { "path": "details.[].expiration", "label": "Approval expires", "format": "date", "params": { "encoding": "timestamp" } },
           { "label": "Sig Deadline", "path": "sigDeadline", "visible": "never" }
@@ -49,7 +57,11 @@
             "path": "permitted.amount",
             "label": "Amount",
             "format": "tokenAmount",
-            "params": { "tokenPath": "permitted.token" },
+            "params": {
+              "tokenPath": "permitted.token",
+              "threshold": "0x8000000000000000000000000000000000000000000000000000000000000000",
+              "message": "Any amount"
+            },
             "visible": "always"
           },
           { "path": "deadline", "label": "Deadline", "format": "date", "params": { "encoding": "timestamp" }, "visible": "always" }
--- a/registry/uniswap/testsv2/eip712-uniswap-permit2.tests.json
+++ b/registry/uniswap/testsv2/eip712-uniswap-permit2.tests.json
@@ -100,6 +100,147 @@
           { "label": "Approval expires", "value": "2026-05-28 20:26:40Z" }
         ]
       }
+    },
+    {
+      "description": "Authorize spending of token - unlimited allowance",
+      "data": {
+        "types": {
+          "EIP712Domain": [
+            { "name": "name", "type": "string" },
+            { "name": "chainId", "type": "uint256" },
+            { "name": "verifyingContract", "type": "address" }
+          ],
+          "PermitDetails": [
+            { "name": "token", "type": "address" },
+            { "name": "amount", "type": "uint160" },
+            { "name": "expiration", "type": "uint48" },
+            { "name": "nonce", "type": "uint48" }
+          ],
+          "PermitSingle": [
+            { "name": "details", "type": "PermitDetails" },
+            { "name": "spender", "type": "address" },
+            { "name": "sigDeadline", "type": "uint256" }
+          ]
+        },
+        "primaryType": "PermitSingle",
+        "domain": { "name": "Permit2", "chainId": 1, "verifyingContract": "0x000000000022D473030F116dDEE9F6B43aC78BA3" },
+        "message": {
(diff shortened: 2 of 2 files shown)


fbwoolf (approved):
Approving. The part that needed checking is the decision to use two different thresholds, and it is right on both counts.

**The `uint160` allowance fields.** `PermitDetails.amount` is `uint160`, and `type(uint160).max` is not merely "a big number" here, it is Permit2's actual sentinel. `AllowanceTransfer._transfer` decrements the stored allowance on every transfer unless it is exactly `type(uint160).max`, so that one value is the only allowance that does not decay. A 2^255 threshold would have been unreachable for a `uint160` field, and any threshold below the max would have labelled a decaying allowance "Unlimited". The declared `0xffffffffffffffffffffffffffffffffffffffff` is 2^160 - 1, and the test message carries exactly that value.

**The `uint256` transfer field.** `PermitTransferFrom.permitted.amount` is a cap on one transfer, not a standing allowance, so 2^255 with "Any amount" is the right pair. Choosing different wording from "Unlimited" is a good call and worth keeping: "Unlimited" on a single-use permit would imply a standing grant that does not exist.

**Verified locally** against `@ethereum-sourcify/clear-signing-test-runner` at the pin this repository uses (`dae3cda`, v0.2.2): 5 cases, 5 pass.

I also probed the boundary, since the whole fix rests on the comparison being inclusive. With `details.amount` set to `type(uint160).max - 1` the field renders as the bare integer `1461501637330902918203684832716283019655932542974`, and at `type(uint160).max` it renders `Unlimited USDC`. So the threshold fires on equality and only on equality, which is exactly the contract's own sentinel.

That result is worth carrying over to #2897. Defect 2 there records that the specification says "above" while every implementation compares `>=`. These two Permit2 fields are the case where the difference is not academic: if the text were ever reconciled in the strict direction, the threshold here would stop firing altogether and the finding this PR fixes would come back silently. The three cases added here now fail loudly if that happens, which is the right guard.

Closes #2677 (1 finding, Medium). Tracked in #2852.

PR #2903, opened 2026-08-18
fix(lifi): render bridge destinationChainId as network name

kuzdogan:
The `_bridgeData.destinationChainId` field in every LI.FI bridge format used `"format": "raw"`, so wallets displayed the bridge destination as a plain chain id number instead of resolving it to a network name.

This changes the format to `"chainId"` on all **72** occurrences, so the field is displayed as the EIP-155 network name (e.g. `Optimism` instead of `10`).

### Details
- Only the `format` value is changed; path, label, and visibility are untouched.
- `chainId` is the correct enum value in `specs/erc7730-v2.schema.json` (const `chainId`, "displayed as a Blockchain explicit name, as defined in EIP-155").
- JSON re-parses clean; the diff is exactly 72 changed lines.

The `interpolatedIntent` strings are left as-is; they reference the same path and will now render the network name inside the intent too.

### DIFF
--- a/registry/lifi/calldata-LIFIDiamond.json
+++ b/registry/lifi/calldata-LIFIDiamond.json
@@ -5974,7 +5974,7 @@
             "params": { "tokenPath": "_bridgeData.sendingAssetId" },
             "visible": "always"
           },
-          { "path": "_bridgeData.destinationChainId", "label": "Destination Chain", "format": "raw", "visible": "always" },
+          { "path": "_bridgeData.destinationChainId", "label": "Destination Chain", "format": "chainId", "visible": "always" },
           {
             "path": "_bridgeData.receiver",
             "label": "Recipient",
@@ -6001,7 +6001,7 @@
             "params": { "tokenPath": "_bridgeData.sendingAssetId" },
             "visible": "always"
           },
-          { "path": "_bridgeData.destinationChainId", "label": "Destination Chain", "format": "raw", "visible": "always" },
+          { "path": "_bridgeData.destinationChainId", "label": "Destination Chain", "format": "chainId", "visible": "always" },
           {
             "path": "_bridgeData.receiver",
             "label": "Recipient",
@@ -6033,7 +6033,7 @@
             "params": { "tokenPath": "_bridgeData.sendingAssetId" },
             "visible": "always"
           },
-          { "path": "_bridgeData.destinationChainId", "label": "Destination Chain", "format": "raw", "visible": "always" },
+          { "path": "_bridgeData.destinationChainId", "label": "Destination Chain", "format": "chainId", "visible": "always" },
           {
             "path": "_bridgeData.receiver",
             "label": "Recipient",
@@ -6060,7 +6060,7 @@
             "params": { "tokenPath": "_bridgeData.sendingAssetId" },
             "visible": "always"
           },
-          { "path": "_bridgeData.destinationChainId", "label": "Destination Chain", "format": "raw", "visible": "always" },
+          { "path": "_bridgeData.destinationChainId", "label": "Destination Chain", "format": "chainId", "visible": "always" },
           {
             "path": "_bridgeData.receiver",
             "label": "Recipient",
@@ -6088,7 +6088,7 @@
             "params": { "tokenPath": "_bridgeData.sendingAssetId" },
             "visible": "always"
           },
-          { "path": "_bridgeData.destinationChainId", "label": "Destination Chain", "format": "raw", "visible": "always" },
+          { "path": "_bridgeData.destinationChainId", "label": "Destination Chain", "format": "chainId", "visible": "always" },
           {
             "path": "_bridgeData.receiver",
             "label": "Recipient",
@@ -6115,7 +6115,7 @@
             "params": { "tokenPath": "_bridgeData.sendingAssetId" },
             "visible": "always"
           },
-          { "path": "_bridgeData.destinationChainId", "label": "Destination Chain", "format": "raw", "visible": "always" },
+          { "path": "_bridgeData.destinationChainId", "label": "Destination Chain", "format": "chainId", "visible": "always" },
           {
             "path": "_bridgeData.receiver",
             "label": "Recipient",
@@ -6146,7 +6146,7 @@
             "params": { "tokenPath": "_bridgeData.sendingAssetId" },
             "visible": "always"
           },
-          { "path": "_bridgeData.destinationChainId", "label": "Destination Chain", "format": "raw", "visible": "always" },
+          { "path": "_bridgeData.destinationChainId", "label": "Destination Chain", "format": "chainId", "visible": "always" },
           {
             "path": "_bridgeData.receiver",
             "label": "Recipient",
@@ -6173,7 +6173,7 @@
             "params": { "tokenPath": "_bridgeData.sendingAssetId" },
             "visible": "always"
           },
-          { "path": "_bridgeData.destinationChainId", "label": "Destination Chain", "format": "raw", "visible": "always" },
+          { "path": "_bridgeData.destinationChainId", "label": "Destination Chain", "format": "chainId", "visible": "always" },
           {
             "path": "_bridgeData.receiver",
             "label": "Recipient",
@@ -6201,7 +6201,7 @@
             "params": { "tokenPath": "_bridgeData.sendingAssetId" },
             "visible": "always"
           },
-          { "path": "_bridgeData.destinationChainId", "label": "Destination Chain", "format": "raw", "visible": "always" },
+          { "path": "_bridgeData.destinationChainId", "label": "Destination Chain", "format": "chainId", "visible": "always" },
           {
             "path": "_bridgeData.receiver",
             "label": "Recipient",
@@ -6229,7 +6229,7 @@
             "params": { "tokenPath": "_bridgeData.sendingAssetId" },
             "visible": "always"
           },
-          { "path": "_bridgeData.destinationChainId", "label": "Destination Chain", "format": "raw", "visible": "always" },
+          { "path": "_bridgeData.destinationChainId", "label": "Destination Chain", "format": "chainId", "visible": "always" },
           {
             "path": "_bridgeData.receiver",
             "label": "Recipient",
@@ -6257,7 +6257,7 @@
             "params": { "tokenPath": "_bridgeData.sendingAssetId" },
             "visible": "always"
           },
-          { "path": "_bridgeData.destinationChainId", "label": "Destination Chain", "format": "raw", "visible": "always" },
+          { "path": "_bridgeData.destinationChainId", "label": "Destination Chain", "format": "chainId", "visible": "always" },
           {
             "path": "_bridgeData.receiver",
             "label": "Recipient",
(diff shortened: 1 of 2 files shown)


kuzdogan:
**Heads-up: these tests cannot pass both reference runners**

This change sets `format: "chainId"` on `_bridgeData.destinationChainId`. The wallet then shows the destination as a network name, not a raw number.

The problem: the spec does not define one source for chain names. Each runner uses its own source, and the two disagree.

- The Sourcify runner reads names from `chainid.network` (`chains_mini.json`). If the id is unknown, it shows the raw number.
- The Rust runner reads names from a small hardcoded table. If the id is unknown, it shows `Chain <id>`.

The two runners produce different text for the same chain id:

| chain id | Sourcify | Rust |
|---|---|---|
| 8453 | `Base` | `Base` |
| 56 | `BNB Smart Chain Mainnet` | `BNB Chain` |
| 4663 | `Robinhood Chain` | `Chain 4663` |
| 1151111081099710 (Solana) | `1151111081099710` | `Chain 1151111081099710` |
| 1885080386571452 (NEAR) | `1885080386571452` | `Chain 1885080386571452` |

The two runners agree on one chain only: Base (8453). CI needs every case to pass on both runners. So no expected value passes both, and this PR cannot go green as it stands.

Two more points:
- Some LI.FI destinations are not EIP-155 chains (Solana, NEAR). The `chainId` format cannot name them.
- The descriptor has 72 `destinationChainId` fields. The test file covers 6 of them.

Suggested path: Publish the fix on the descriptor as is, but think about the tests fix in a separate issue.

0xDEnYO:
Thanks for picking this up @kuzdogan — the goal is right and it's actually our own change: we switched `_bridgeData.destinationChainId` from `raw` to `chainId` on our side and it's sitting in our open sync PR #2901. **But it turns out the blanket switch is wrong for a small subset of our bridges, and we'd like to ask you to close this PR in favour of #2901.** Details below.

### Why not all 72

One of our internal chain ids collides with a registered EIP-155 id. `LIFI_CHAIN_ID_HYPERCORE` is **1337**, and `chainid.network/chains_mini.json` — the exact source the Sourcify reference runner uses — resolves it to **`Geth Testnet`**:

```
1337 -> ["Geth Testnet"]
```

So with `format: "chainId"`, a user bridging to Hypercore gets `Destination Chain: Geth Testnet` on their hardware wallet. A confidently-wrong network name on a signing screen is worse than a bare number. `UnitFacet` is the worst case — it is Hypercore-only, so it would name the wrong chain on *every single* transfer.

Every other LI.FI-internal id (Solana, BTC, Sui, Stellar, Aptos, LTC/BCH/DOGE, Tron) is far outside the allocated range and simply unregistered, so it degrades to digits — harmless, and `chainId` is strictly better there. Verified against the same file:

```
1151111081099710 -> []      # Solana
1885080386571452 -> []      # NEAR
```

Worth noting this is the same weakness you flagged separately — that ERC-7730 doesn't define a canonical chain-name source and the two reference runners disagree. We're working around it from our side rather than waiting on a spec answer.

### How we're fixing it

Our generator (`tasks/buildClearSigningProposal.ts` in `lifinance/contracts`) now picks the format **per bridge** instead of using one shared constant: bridges that can reach a colliding id keep `format: "raw"`, everything else gets `"chainId"`. That's **8 of 72** entries — Mayan, Unit, PolymerCCTP, AcrossV4Swap — and the affected set is derived from our facet sources rather than hardcoded, with two generator-level guards:

1. A new internal chain id that lands on a reserved EIP-155 id fails the build unless explicitly acknowledged.
2. A bridge that reaches a colliding id but matches no emitted format (facet rename / naming drift) fails the build, so the `raw` fallback can't be silently dropped.

PR: https://github.com/lifinance/contracts/pull/2225

### Why close rather than merge

`registry/lifi/calldata-LIFIDiamond.json` is generated on our side and pushed here by an automated sync workflow. It rebases `sync/lifi-erc7730` onto `master`, regenerates the descriptor, `--force-with-lease` pushes, and **updates the existing #2901 in place**. So a hand edit to that file gets overwritten on our next sync regardless of merge order — and if this merges first, #2901 would silently revert it. Once #2225 lands on our `main` the sync fires automatically and **#2901 will be updated shortly** with the corrected per-bridge formats. One PR, correct end state.

Two smaller notes while we're here:

- **Test fixtures**: your `testsv2` update looks correct and is neutral w.r.t. our fix — the only fixture touching an affected bridge is the Mayan one, and it uses `1151111081099710`, which renders as digits under both formats. Note there's no 1337/Hypercore or Unit fixture, which is why the suite stays green while the bug ships. We don't currently manage `testsv2/` in our sync — happy to add a Hypercore fixture in a follow-up if you'd like it, just let us know whether you'd prefer that from us or upstream.
- Some entries this PR touches (`CBridge`, `HopL1`/`HopL2`) are for facets we've since deprecated and removed from our diamond. Our sync merges `display.formats` additively and never prunes, so those are stale here. We'll clean them up separately.

Happy to be told you'd rather land this and let our sync correct it after — but closing keeps the history cleaner and avoids shipping the wrong chain name in the interim. Thanks again for flagging it.

0xDEnYO:
Thanks for picking this up @kuzdogan — and apologies for the crossed wires, this descriptor is auto-synced from `lifinance/contracts` so we had the same change in flight from our side.

Your PR turned out to be the most useful piece of evidence in the whole investigation, so it is worth writing down what it showed before closing it.

**`format: chainId` cannot be made to pass, in either direction.** Your PR tunes the fixture to the Sourcify implementation (`Robinhood Chain`, `BNB Smart Chain Mainnet`) — Sourcify passes, the Rust runner fails. Our sync PR #2901 leaves the fixture expecting raw numbers — both runners fail. Same descriptor, opposite fixtures, and there is no third option, because the two implementations disagree on three separate axes:

| destinationChainId | `llbartekll/clear-signing` | `sourcifyeth/clear-signing-test-runner` |
| --- | --- | --- |
| `8453` | `Base` | `Base` |
| `56` | `BNB Chain` | `BNB Smart Chain Mainnet` |
| `4663` | `Chain 4663` | `Robinhood Chain` |
| `1151111081099710` | `Chain 1151111081099710` | `1151111081099710` |

The last row is the important one. `_bridgeData.destinationChainId` is **not** an EIP-155 id space for us: `src/Helpers/LiFiData.sol` mints synthetic ids for every non-EVM destination — Solana `1151111081099710`, Tron `1885080386571452`, Stellar, Sui, Aptos, the Bitcoin family, HyperCore `1337`. `specs/erc-7730.md:1411-1416` defines `chainId` as *"converted to a Blockchain name using EIP-155 reference values"* and, unlike `tokenAmount` or `nft`, specifies **no fallback** — so for those ids the rendering is genuinely implementation-defined and the runners are both behaving correctly.

Which also means it was never only a CI problem. A real wallet has no fixture, so a user bridging to Solana would sign a screen reading *"to chain Chain 1151111081099710"*. Note your PR's fixture leaves the Solana and Tron cases as bare numbers, which is exactly the divergence surfacing.

**We have opened #2914 instead**, using `format: enum` against a LI.FI-authored `$.metadata.enums.lifiChains`. The rendered string becomes a function of the descriptor alone — identical on every implementation, no network lookup, and correct for non-EVM destinations. It follows the precedent already in the registry for a project-specific non-EIP-155 destination id space (`threshold`'s `wormholeChain`, including `"21": "Sui"`; `hyperliquid`'s `cctpDomains`). Descriptor and `testsv2` move in the same commit, and both pinned runners pass 15/15 locally, including a case whose destination is deliberately absent from the enum so the decimal fallback is regression-covered.

So **could we close this one in favour of #2914?** As far as we can tell it covers everything here plus the non-EVM cases — but you know the Sourcify side far better than we do, so if there is anything in your version we have missed, say the word and we will fold it in rather than close.

One thing from your PR we did keep: it made us notice that the interpolated intents read `"… to chain {destinationChainId} …"`, which renders as *"to chain BNB Chain"* / *"to chain Chain 4663"*. That is fixed in #2914 too.

There is also an open question on #2914 that would benefit from your eyes: `erc7730 calldata` rejects enum keys above 255 (`le=255` in the v1 converter), so we cannot tell whether a chain-id-keyed enum works on an actual Ledger device. Details in the comment there.

0xDEnYO:
Correction to my comment above, before anyone acts on it: **please hold off closing this.** I asked you to close in favour of #2914, and #2914 has since turned out not to work on a Ledger.

The ENUM_VALUE struct encodes an enum entry's value as a single byte (`obj.value.to_bytes(1)` in `erc7730/convert/calldata/v1/tlv.py:255`, model bound `le=255`), and `convert_enums` takes that value from the enum **key** — the chain id. So no chain id above 255 can be represented, and `erc7730 calldata` drops the whole LI.FI descriptor to `[]` while exiting 0. `threshold`'s `wormholeChain` and `hyperliquid`'s `cctpDomains` convert fine only because their ids happen to be ≤ 255, so they were not the precedent I claimed they were.

Everything in my previous comment about **`chainId` itself** still stands, and this PR is still evidence for it — your fixture tuned to Sourcify passes Sourcify and fails Rust, ours expecting raw numbers fails both, and `specs/erc-7730.md:1411-1416` defines `chainId` over EIP-155 values with no fallback while ~10 of LI.FI's destinations are non-EVM synthetic ids. So I do not think this PR can go green as it is either.

Where that leaves things: `raw` is the only one of the three variants that both converts on-device and passes both runners, so we will probably revert to that in the short term, and the real fix is upstream — most likely defining a fallback for `chainId` the way the spec already does for `tokenAmount` and `nft`. Details and measurements on #2914; would value your read on it given you know the Sourcify side.

Apologies for the noise on your PR.

0xDEnYO:
Final update from our side, and it changes the conclusion for this PR too — sharing the measurements because they point at a data-quality issue in the resolution path rather than anything about the descriptor.

We probed 30 chain ids through both reference implementations to see what `chainId` actually renders. **Only 7 render identically:** `100`, `8453`, `42161`, `42170`, `59144`, `534352`, `7777777`. Most of the rest are cosmetic (`1` → "Ethereum Mainnet" vs "Ethereum", `10` → "OP Mainnet" vs "Optimism", `137` → "Polygon Mainnet" vs "Polygon"). But two are not cosmetic:

| chain id | LI.FI destination | `chains_mini.json` name |
| --- | --- | --- |
| `999` | **HyperEVM** (mainnet) | **Wanchain Testnet** |
| `1337` | **HyperCore** (mainnet) | **Geth Testnet** |

Both are live mainnet bridge destinations that would be displayed to a signer as *testnets*. Verified directly against the source:

```console
$ curl -s https://chainid.network/chains_mini.json | jq '.[] | select(.chainId==999 or .chainId==1337) | {chainId, name}'
{ "chainId": 999,  "name": "Wanchain Testnet" }
{ "chainId": 1337, "name": "Geth Testnet" }
```

That is worth knowing on the Sourcify side independently of ERC-7730: `src/chain-info.ts` resolves names live from that list, so any descriptor using `format: chainId` inherits whatever it says, including stale or ambiguous entries for ids that have been reused. `747474` also comes back as lowercase `"katana"`.

So **we are reverting to `raw`**, which means this PR's change should not land either — I do not think `chainId` is safe for this field until the spec pins a name source and a fallback. **Could you close this one?** Our sync PR #2901 will go green on its own once our revert merges (its descriptor regenerates as `raw` and matches the fixture already on `master` — verified locally, both runners 14/14), so nothing needs to be merged here at all.

Thanks for the patience through the back-and-forth on this — your PR was genuinely the clearest single piece of evidence that `chainId` could not be made to pass, since it demonstrated the Sourcify-tuned direction while ours demonstrated the other.

The two underlying gaps are filed separately so they do not sit in a closed PR.

kuzdogan:
Closing in favor of deffering to implementer's own definition at #2901.

The wider issue of chainId and chain name mismatches are discussed and documented in #2904

PR #2905, opened 2026-08-19
ci: automate the mechanical parts of the descriptor review

manuelwedler:
The purpose of this pull request is to automate as much as possible of the manual
[review checklist](https://github.com/sourcifyeth/clear-signing-docs/blob/8e70274/docs/registry/reviewing-prs.md).

## Description

- Every pull request runs the clear signing tests, from a fork too. The `run-tests` label is gone.
- A fork run cannot write a comment, so two new workflows do it from the base repository. One posts a note when a commit arrives. The other replaces it with the results.
- Each test job now fails on its own cases, so a red check names the runner and the descriptor.
- Every affected descriptor must have a test file. One descriptor with a test file no longer hides another without one.
- A second comment names the descriptors that have no `interpolatedIntent`, and the two deprecated context keys. It never blocks the pull request.
- A new check enforces the file names in the registry. It renames one file that already broke them.
- A new check rejects a change to a descriptor that an auditor already attested, and follows `includes` to catch a shared file that changes an attested descriptor.

## Why the label can go

The Sourcify runner and the Rust runner consume no secrets. So a plain `pull_request` run is sufficient, and that event fires for a fork.

A fork run does get a read-only token. It cannot write a comment and it cannot set a commit status. Only that part needed to move.

## What this replaces

A cross-repo pull request ran the tests only after a maintainer added the `run-tests` label.

- Before the label, the `pull_request` run reported a column of skipped checks. #2597 shows this: four runs, one with every test job skipped.
- After the label, the `pull_request_target` run reported against the base branch. A later push did not invalidate its results, so the pull request kept a verdict for an earlier commit.

## The three workflows

| File | Trigger | Task |
| --- | --- | --- |
| `clear-signing-tests.yml` | `pull_request` | Runs the tests. Gives the red or green verdict. |
| `clear-signing-tests-started.yml` | `pull_request_target` | Posts "tests are queued" for the new commit. |
| `clear-signing-tests-results.yml` | `workflow_run` | Posts the results table and the recommendations. |

Each file holds only the jobs of its own trigger, so no job shows as a skipped check.

`notify-fork-pr.yml` is deleted. It only told a contributor to wait for the label.

## Where the verdict comes from

`post-implementation-results` held the aggregate failure. It now runs from `workflow_run`, and a `workflow_run` job cannot fail a pull request.

So `upload-test-results` fails its own job when a case does not pass. Each runner and each descriptor gets its own red check, with an annotation on the test file. A lost comment therefore cannot make a failed pull request look green.

## A test file for every descriptor

`detect-testsv2` reported `has_tests` as true as soon as **one** affected descriptor had a test file, and `require-testsv2` only ran when that flag was false. A pull request with two descriptors, one tested and one not, passed silently.

`list-affected-descriptors.js` now also returns `missing_tests`. `require-testsv2` keys off that list. The results comment names the same descriptors above the table:

```
❌ 1 affected descriptor(s) have no test file:

- `registry/celo/calldata-celo_accounts.json`

Create a test file at `registry/<entity>/testsv2/<descriptor-name>.tests.json`. See …

| Entity | Descriptor | Case | @ethereum-sourcify/clear-signing@0.2.2 |
| --- | --- | --- | :---: |
| weth | calldata-weth | Wrap 1 ETH | ✅ |
```

The tests of the descriptors that do have a test file still run and still report.

## Recommended and deprecated fields

ERC-7730 recommends `interpolatedIntent` and deprecates `context.contract.abi` and `context.eip712.schemas`. None of them is an error, so none belongs in a blocking check. But a contributor never heard about them.

`check-recommended-fields.js` names them for the descriptors that a pull request adds or changes. `detect-testsv2` runs it and puts the output in the `pr-context` artifact; the results workflow posts it.

```
## Clear signing recommendations

These are suggestions. They do not block this pull request.

**1 format(s) have no `interpolatedIntent`.** A wallet prefers it over `intent`, because it
puts the values of the transaction in the sentence that the signer reads.

- `registry/weth/calldata-weth.json` — `deposit()`
```

It is a second comment, under its own marker. The two have different lifetimes: the results are replaced on every push, while this one is removed as soon as the author adds the fields.

## File names in the registry

A prefix says what a file is: `calldata-` and `eip712-` mark a root descriptor, and `common-` marks a shared file that they include. Nothing enforced it, and one file already broke it.

`validate_file_names` walks every entity folder, steps over `tests/`, `testsv2/` and `sigs/`, and annotates each file that does not match. It runs over the whole registry, not only the changed files, so it needs no changed-files handling.

**This pull request therefore renames one descriptor file**, which is the only violation in the registry:

```
registry/uniswap/uniswap-common-eip712.json  ->  registry/uniswap/common-eip712-uniswap.json
```

The five uniswap descriptors that include it are updated. The rename is safe: `registry/uniswap/sigs/` does not exist, so no attestation covers a file whose canonical JSON the rename changes. `generate-index.js --check` still exits 0, because a shared file is never indexed.

The README gains the `sigs/` and `tests/` subfolders, the real `ercs/` file names, and `erc7730-v2.schema.json`, which was missing while the v1 schema was labelled as the latest. `specs/README.md` gains the tests-v2 schema and a note that the attestations deliberately have no schema.

## The recommendations step skips on this pull request

`check-recommended-fields.js` runs from `base-scripts`, a checkout of the base ref, because only base-controlled code should grade a pull request. This pull request adds that script, so the base ref does not hold it yet.

The step therefore carries a guard:

```yaml
if: always() && hashFiles('base-scripts/.github/scripts/check-recommended-fields.js') != ''
```

It skips here, and starts to work by itself after the merge, when the script reaches `master`. Without the guard the step would fail, and because `Detect testsv2 files` builds the matrix, the whole test suite would skip on this pull request — the very thing under review would go unexercised.

The consequence is only that no recommendations comment appears on this pull request. The results workflow treats the missing file as "nothing to report".

## Attested descriptors

An auditor signs the exact content of a descriptor, so any change makes the attestation false. The [review guideline](https://github.com/sourcifyeth/clear-signing-docs/blob/8e70274/docs/registry/reviewing-prs.md) therefore rejects such a change and asks for a new descriptor file, whose name carries the version. `validate_attested_descriptors` enforces that with two checks.

**A change to an attested descriptor.** It runs `list-affected-descriptors.js`, so it uses the affected set and not the changed set. A `common-*.json` file and an `ercs/` template are inlined through `includes`, so an edit there silently changes every descriptor that includes it, attested ones too. The guideline names this case and says no check catches it.

| Change | Attested descriptors reached |
| --- | --- |
| `registry/lido/calldata-stETH.json` | 1 |
| `registry/kiln/common-KilnVaults.json` | 27 |
| `ercs/calldata-erc4626-vaults.json` | 131, two `includes` levels deep |
| `ercs/eip712-erc2612-permit.json` | 0 — its 74 descriptors are not attested |

**A delete or a rename.** The second check fails when an attestation has no descriptor. It looks from the attestation side, because the file name is the only bond between the two, and because `tj-actions` does not report the old path of a rename. Adding a new descriptor beside the attested one still passes, which is the path the guideline asks for.

Both write their total to the step summary, since GitHub renders only ten annotations per step.

## Security notes

- The results workflow takes the pull request from the `workflow_run` event, never from the artifact. A fork writes that artifact, so a value inside it must not select the pull request to comment on.
- Neither privileged workflow checks out or runs any code from the pull request head. Both read it only as data.
- Values that come from a runner or from a descriptor are escaped before they enter a comment. GitHub permits some HTML there, so a case description or a format selector could otherwise forge a passing table.
- Both comment bodies are capped below the 65536-character limit of the API.
- **A `pull_request` run uses the workflow file of the merge commit, so a fork can change these jobs in its own pull request.** The same is already true of `pull_request.yml`. Read the diff of a pull request that touches `.github/` before you trust its checks. `CLAUDE.md` says so too.

## One thing to know

GitHub asks a maintainer to approve the first workflow run of a new contributor. That is a repository setting, not a file, and it already applies to `pull_request.yml` today. It is one click instead of a label.

Until that click the tests do not start, and the queued note stays on the pull request. The note says so, so a contributor is not left guessing.

## Verification

`workflow_run` and `pull_request_target` only take effect from the default branch, so this pull request cannot exercise its own comment jobs. Checked locally instead:

- All YAML parses. Every `github-script` body passes `node --check`. Every `run:` block passes `bash -n`.
- The `pr-context` step produces the right `context.json`, test files and recommendations, and also after a failed detect step, where the outputs are empty.
- The `upload-test-results` gate exits 0 for `pass` and 1 for `fail`, `error` and `skipped`, which matches the previous rule.
- The comment script was run against prepared artifacts, with stubs for the API. It produced the bodies shown above, including the mixed case where one descriptor has tests and another does not.
- A hostile format selector, with a backtick and HTML, stays inside its code span.
- The file name check exits 0 on the registry as it stands, and flags a stray file at the top level of an entity folder, one in an unexpected subfolder, one three levels deep, and a path that holds a space. It ignores the three known subfolders.
- After the rename the five includers still resolve, and both `generate-index.js --validate` and `--check` exit 0.
- The attestation checks pass on the registry as it stands, with its 181 attestations, and fail for a changed attested descriptor, a shared file that reaches one, a deleted one and a renamed one. A broken detection script fails the check instead of passing it.


kuzdogan (dismissed):
Approving. The workflows, the scripts, and the checks all do what the description says. I ran the file-name check, the attestation scan, the index build, and `list-affected-descriptors.js` against the registry, and each one passed.

One non-blocking note for a follow-up:

`check-recommended-fields.js` reads the raw changed files, not the affected set. So a change to a shared `common-` file does not raise the recommendation on the descriptors that include it. Every other check here resolves `includes`; this step is the one that does not.

The fix is small. Pass `steps.detect.outputs.affected_descriptors` to the script instead of `all_changed_files`. The script already takes file paths as arguments, so no code change is needed there.

Decide the intent first: if a shared-file change should nag its includers, apply the fix; if the comment should stay scoped to a descriptor's own file, keep the current behavior and say so in the docstring.

manuelwedler:
Thanks — the gap is real, and it is wider than the note says: a shared file was skipped entirely, not only its includers. `registry/1inch/common-AggregationRouterV6.json` holds 21 formats without an `interpolatedIntent` and produced nothing.

I took the other fix, because I measured both. For a change to `ercs/calldata-erc4626-vaults.json`:

| Approach | Result |
| --- | --- |
| Affected set, files read as written | 0 findings — the 133 includers declare no formats of their own |
| Affected set, with `includes` resolved | 532 findings over 133 files, the same 4 formats repeated |
| Read the changed shared file (`72aab1b`) | 4 findings, on the file that declares them |

A format belongs to the file that declares it. `registry/kiln/calldata-Vault-USDC-Aave-v3.json` has zero formats of its own, so annotating it would name a file where the fix cannot be made.

So the script now reads `common-*.json` and `ercs/*.json` as well, at any depth, and still does not resolve `includes`. The docstring says why.

kuzdogan (approved):

manuelwedler:
@kuzdogan we should ask all open descriptor PRs to rebase on master to apply the CI changes.

PR #2906, opened 2026-08-19
Lido – cover L2 deployments

alx-khramov:
## Changes
- Adds wstETH for L2 chain ids 10, 56, 130, 8453, 42161 and 59144 with ERC20 methods
- Adds stETH for L2 chain ids 10 and 130 with ERC20 + wrap/unwrap methods


Jeday:
LGTM

tamtamchik (approved):

manuelwedler:

fbwoolf:
Two blockers. Happy to look at the rest once these are in.

**1. `validate attested descriptors` is red.** `registry/lido/calldata-stETH.json` is attested in
`registry/lido/sigs/calldata-stETH.eip155-1-0x3846c3A30E62075Fa916216b35EF04B8F53931f6.json`, so it
cannot be edited in place. The three added L1 formats (`transferFrom`, `transferShares`,
`transferSharesFrom`) need to move to a new descriptor file.

**2. The wrap/unwrap intents in `calldata-stETH-L2.json` are inverted.** On L2 the roles are the
reverse of L1: stETH is the wrapper and wstETH is the wrappable token. `TOKEN_TO_WRAP_FROM()` on the
stETH deployments returns `0x1f32b1c2345538c0c6f582fcb022739c4a194ebb` (chain 10) and
`0xc02fe7317d4eb8753a02c35fe019786854a92001` (chain 130), the wstETH addresses this PR adds.

| format key | what the call does | current intent | should read |
| --- | --- | --- | --- |
| `unwrap(uint256 stETHAmount)` | spends stETH, returns wstETH | `Wrap stETH` / `Wrap {stETHAmount}` | `Unwrap stETH to wstETH` |
| `wrap(uint256 wstETHAmount)` | spends wstETH, returns stETH | `Unwrap wstETH to stETH` / `Unwrap {wstETHAmount}` | `Wrap wstETH to stETH` |
| `unwrapShares(uint256 wstETHAmount)` | spends stETH, returns wstETH | `Wrap stETH` / `Wrap stETH for {wstETHAmount}` | `Unwrap stETH for {wstETHAmount}` |

The intents look carried over from L1 `calldata-wstETH.json`, where the direction is opposite. A
signer sees "Wrap stETH" for a call that unwraps. The test descriptions carry the same inversion
(`"Wrap stETH - chain 10"` is calldata `0xde0e9a3e` = `unwrap`; `"Unwrap wstETH to stETH - chain 130"`
is `0xea598cb0` = `wrap`), so they need the same fix. Field labels and amounts are correct as they
stand.

Worth noting the tests cannot catch this: the runner checks rendering against the descriptor, so an
inverted intent passes.

Jeday:
Hey @fbwoolf!
We will move new formats to separate file if so required.

As per wrap inversion, this is intentional:
 - the onchain naming corresponds to contract logic because for L2 wsteth is bridged token that gets `wrapped`
 - but on our UI and with naming of the token the inversion corresponds correctly, user can wrap stETH to (wrapped)stETH and unwrap it back on both mainnet and L2.

fbwoolf:
Yes, a new file is needed. `calldata-stETH.json` is attested in
`registry/lido/sigs/calldata-stETH.eip155-1-0x3846c3A30E62075Fa916216b35EF04B8F53931f6.json`, so
`validate attested descriptors` fails on any edit to it. Putting the added formats in their own
descriptor keeps that attestation valid and turns the check green.

I'll approve and merge it once the failing test passes.

Jeday:
@fbwoolf Hey! I checked the proposed change locally. But seems that repo rules prohibit having two descriptor files for 1 address.
[Here is the check that fails](https://github.com/ethereum/clear-signing-erc7730-registry/blob/master/tools/scripts/generate-index.js#L101-L109)

I am not sure how best to proceed along the repo rules.

Could the updated mainnet stETH and wstETH descriptors be re-attested? The goal of this change is to add clear signing to critical methods that modify token balances.

manuelwedler:
Hey @Jeday, yes this is a limitation of the current index structure. The problem is that the index doesn't include attestations nor spec version in the mapping. I'm aware of this problem and we should resolve this soon. I just created this issue to keep track of it: https://github.com/ethereum/clear-signing-erc7730-registry/issues/2959

For now, I think the easiest way to move forward for you is to remove the attestation files for the descriptor in a separate PR and then change the existing descriptor file in here (not add a new file).

Jeday:
@manuelwedler Alright, I reverted L1 changes and will bring it other PR. CI all green. 
What's the process on getting attestations for L2s and also withdrawal queue descriptors signed?

fbwoolf (approved):

Issue #2459, opened 2026-03-18
LedgerHQ/coin-apps is private


PatrickAlphaC:
Would be nice to be able to see what it looks like on the device

PatrickAlphaC:
This issue is still valid. Output of running the setup for testing:

```
% ./setup.sh 
============================================
Clear Signing Tester - Local Setup
============================================

Checking prerequisites...
✅ All prerequisites found

📦 Cloning device-sdk-ts...
Cloning into 'dmk'...
remote: Enumerating objects: 3686, done.
remote: Counting objects: 100% (3686/3686), done.
remote: Compressing objects: 100% (3070/3070), done.
remote: Total 3686 (delta 780), reused 2325 (delta 441), pack-reused 0 (from 0)
Receiving objects: 100% (3686/3686), 28.11 MiB | 9.13 MiB/s, done.
Resolving deltas: 100% (780/780), done.

📦 Cloning coin-apps (private repo - requires access)...
   This may prompt for authentication...
Cloning into 'coin-apps'...
remote: Repository not found.
fatal: repository 'https://github.com/LedgerHQ/coin-apps.git/' not found
```

PatrickAlphaC:
After speaking with Ledger, it may be better/easier to just use a different test suite for the interim, and make the current test docs "legacy" or "ledger optional". They have plans to open source the binaries of those repos so that people can run this portion of the tests in the future.

PR #2914, opened 2026-08-24
fix(lifi): render destinationChainId via a LI.FI-authored enum instead of format: chainId

0xDEnYO:
## Problem

`registry/lifi/calldata-LIFIDiamond.json` renders `_bridgeData.destinationChainId`, and LI.FI's value space for that field is **not** EIP-155. Alongside ordinary chain ids it carries LI.FI-internal synthetic ids for non-EVM destinations (`src/Helpers/LiFiData.sol` in `lifinance/contracts`):

| id | destination |
|---|---|
| `1151111081099710` | Solana |
| `1885080386571452` | Tron |
| `1201081091099710` | Stellar |
| `9270000000000000` | Sui |
| `9271000000000010` | Aptos |
| `20000000000001…4` | Bitcoin, Bitcoin Cash, Litecoin, Dogecoin |
| `1337` | HyperCore |

`format: chainId` is defined as *"Value is converted to a Blockchain name using EIP-155 reference values"* (`specs/erc-7730.md:1411-1416`) and — unlike `tokenAmount` or `nft` — **defines no fallback** for a value it cannot resolve. So for the ids above the rendering is implementation-defined, and the two runners this repo pins in CI genuinely diverge:

| destinationChainId | `llbartekll/clear-signing` | `sourcifyeth/clear-signing-test-runner` |
|---|---|---|
| `8453` | `Base` | `Base` |
| `56` | `BNB Chain` | `BNB Smart Chain Mainnet` |
| `4663` | `Chain 4663` | `Robinhood Chain` |
| `1151111081099710` | `Chain 1151111081099710` | `1151111081099710` |

They disagree on the resolved name, on unknown-but-EIP-155 ids, **and on the fallback shape itself**. `#2901` (fixture expects raw numbers) fails both runners; `#2903` (fixture tuned to Sourcify) passes Sourcify and fails Rust. No single fixture can satisfy both — and more importantly, a real wallet has no fixture at all, so a user bridging to Solana signs a screen reading *"to chain Chain 1151111081099710"*.

## Fix

Render the field with `format: enum` against a LI.FI-authored `$.metadata.enums.lifiChains`. The displayed string becomes a function of the descriptor alone: identical on every implementation, no network lookup, no undefined behaviour, and correct for non-EVM destinations.

This is the existing registry precedent for exactly this situation — a project-specific, non-EIP-155 destination id space including non-EVM members:

- `registry/threshold/calldata-L2WormholeGateway.json` → `$.metadata.enums.wormholeChain` (`"21": "Sui"`)
- `registry/hyperliquid/calldata-CctpExtension.json` → `$.metadata.enums.cctpDomains`

Also drops the literal word "chain" from the interpolated intents — `"… to chain {_bridgeData.destinationChainId} …"` otherwise renders as *"to chain BNB Chain"* / *"to chain Chain 4663"*.

## Note on the 32-entry cap

The enum holds **32** entries, not the ~80 destinations LI.FI actually bridges to, because `erc7730 lint` — which `validate descriptors` runs — rejects a larger one:

```text
Invalid value at $.metadata.enums.lifiChains: Dictionary should have at most 32 items after validation, not 81
```

That cap lives in the linter's `EnumDefinition` model (`max_length=32`); **`specs/erc7730-v2.schema.json` in this repo sets no limit** on `$metadata/enums` entries, and `check-jsonschema` accepts the 81-entry version happily. That looks like an unintended tool-vs-spec divergence, and it is a real constraint for bridge aggregators — we would raise the cap if you are open to it (happy to open a PR against `python-erc7730`).

Until then the 32 cover every non-EVM destination — the ids a user has no chance of recognising — plus the EVM chains with the broadest route support. An unlisted id renders as its decimal number, identically on both runners and no worse than the `raw` format this replaces.

## Verification

Run locally against both runners at the exact SHAs this repo pins in `.github/actions/run-*-tests/action.yml`:

| runner | ref | result |
|---|---|---|
| `sourcifyeth/clear-signing-test-runner` | `dae3cda` | **14/14 pass** |
| `llbartekll/clear-signing` | `10605ba` | **14/14 pass** |

Plus `erc7730 lint` (exit 0), `erc7730 format` (no-op on this file), and `check-jsonschema` against `specs/erc7730-v2.schema.json` and `specs/erc7730-tests-v2.schema.json`.

Both the descriptor and the `testsv2` expectations are updated **in the same commit** so the two are never inconsistent on `master`.

## Relationship to the open LI.FI PRs

- **#2901** (`sync/lifi-erc7730`, our sync bot) — currently red on both runners. Superseded by this PR for the `destinationChainId` question; the bot will re-sync ABI/deployments separately once this lands.
- **#2903** — thanks for taking this on. It makes the same `raw → chainId` change and tunes the fixture to one implementation, which is what surfaced the divergence so clearly. Superseded by this PR; commenting there separately.

## Provenance

Both files are generated from `lifinance/contracts` (`tasks/buildClearSigningProposal.ts` → `tasks/generateLedgerClearSigning.ts`), so the enum stays in sync with LI.FI's chain list rather than being hand-maintained here. The corresponding `lifinance/contracts` PR: https://github.com/lifinance/contracts/pull/2260


0xDEnYO:
One open question we could not answer from outside, and it is the one thing that could invalidate this approach — flagging it up front rather than after a merge.

`erc7730 calldata`, the converter that produces the on-device v1 descriptor, rejects this enum:

```text
$ erc7730 calldata registry/lifi/calldata-LIFIDiamond.json
🟠 warning: Error processing v2 ERC-7730 file None, skipping it
🔴 error: Invalid value at $.value: Value "999" is not valid: Input should be less than or equal to 255
[]   (exit 0)
```

`convert/calldata/v1/enum.py` does `value=int(ordinal)` into `CalldataDescriptorInstructionEnumValueV1.value`, which is constrained `ge=0, le=255`. Every chain id is above that, so **no** chain-id-keyed enum can round-trip through that path — including `registry/threshold/calldata-L2WormholeGateway.json`'s `wormholeChain` (`"21": "Sui"` fits, but the concept is the same) once ids grow.

Neither `validate descriptors` nor the `testsv2` jobs run `erc7730 calldata`, and the Ledger Speculos jobs are disabled (#2459), so nothing in CI catches this — and note the command exits 0 while emitting the error.

**Question:** is a v2 `metadata.enums` keyed by values above 255 expected to work on a Ledger device, or is the `le=255` bound in the converter the real constraint? If it is the real constraint, then chain-id-keyed enums are not viable on-device and we would rather know now — our fallback is to keep `raw`, since `chainId` is what this PR set out to avoid.

Two smaller observations from the same pass, both incidental:

1. **`EnumDefinition` is capped at 32 entries in the linter but not in the spec.** `erc7730` `model/metadata.py:120` sets `max_length=32` (with a `# TODO refine`), while `specs/erc7730-v2.schema.json` has no `maxProperties` on `$metadata/enums` at all — `check-jsonschema` accepts an 81-entry map that `erc7730 lint` rejects. For a bridge aggregator that cap is the binding constraint (we bridge to ~80 destinations and had to cut to 32). Happy to open a PR against `python-erc7730` raising it if you agree it is unintended.
2. **`pip install erc7730` is unpinned** in `validate_descriptors`, `validate_attested_descriptors` and `format.yml`, so CI silently follows the latest release. Today latest == what we tested (1.0.10), so there is no drift, but a linter change can turn every open PR red with no commit in this repo.

0xDEnYO:
**Converting to draft — please do not merge yet.** I asked the enum-key question above and then went and answered it myself, and the answer is bad enough that this PR should not land as it stands.

**A chain-id-keyed enum cannot be represented on-device.** `tlv_enum_value` in `erc7730/convert/calldata/v1/tlv.py:255` encodes the ENUM_VALUE entry as:

```python
out += tlv(CalldataDescriptorEnumValueTag.VALUE, obj.value.to_bytes(1))   # one byte
```

next to `obj.chain_id.to_bytes(8)` two lines above, so the width looks deliberate rather than accidental — and `CalldataDescriptorInstructionEnumValueV1.value` is correspondingly `ge=0, le=255`. `convert_enums` sets `value=int(ordinal)` from the enum **key**, i.e. the chain id. So this is the wire format, not a linter bound, and no chain id above 255 fits.

Measured on this branch versus `master`:

| descriptor variant | `erc7730 calldata` output | `testsv2` (both runners) |
| --- | --- | --- |
| `raw` (current `master`) | 25,382,949 bytes ✅ | green |
| `chainId` (#2901, #2903) | 25,385,397 bytes ✅ | **fails both** |
| `enum` (this PR) | **`[]` — descriptor dropped entirely** ❌ | green |

So this PR is the only one of the three that takes `calldata-LIFIDiamond` from "produces an on-device descriptor" to "produces nothing at all". On a Ledger that is not a worse label, it is **no clear signing for the LI.FI diamond**, which is far worse than the bare number we were trying to improve on. And it fails silently — `erc7730 calldata` exits 0 while emitting that error, and neither `lint` nor `testsv2` touches this path (the Speculos jobs being disabled, #2459).

Worth noting for the record: `registry/threshold/calldata-L2WormholeGateway.json` and `registry/hyperliquid/calldata-CctpExtension.json` both convert fine — because Wormhole chain ids and CCTP domain ids are all ≤ 255. I cited them as precedent for a project-specific destination-id enum, and that precedent does not transfer to EIP-155 chain ids. My mistake, and the one-byte encoding is why.

That leaves us without a good option, which is really the point of this comment:

- **`chainId`** — converts on-device, but the spec (`specs/erc-7730.md:1411-1416`) defines it only over EIP-155 reference values with no fallback, so the two `testsv2` runners diverge, and ~10 of our destinations are non-EVM ids that no EIP-155 lookup can resolve.
- **`enum`** — renders identically in software wallets, unrepresentable on Ledger.
- **`raw`** — works everywhere, deterministic, and shows the user a bare number.

We will most likely revert to `raw` for now, and I would rather spend the conversation on one of the two underlying gaps than on this PR:

1. **Define a fallback for `chainId`.** The spec defines one for `tokenAmount` ("raw value plus an Unknown token warning") and for `nft`, but not for `chainId`. If it said "render the decimal id when the chain is unknown", `chainId` would be deterministic across implementations, the `testsv2` divergence would disappear, and it would be honest about non-EIP-155 ids. This looks like the smallest correct fix, and it is a spec change rather than an implementation one.
2. **Widen ENUM_VALUE, or say plainly that enums are for small ordinals.** If a new ENUM_VALUE struct version carried a wider value, chain-id enums would work on-device. If that is not wanted, then the enum primitive is only for small ordinals and the schema/spec should say so — right now nothing warns you, and `erc7730 lint` happily passes a descriptor that `erc7730 calldata` silently discards.

Happy to do the work on either if you tell me which direction you would accept. Flagging separately that a `lint`-clean descriptor being dropped by `calldata` with exit code 0 seems worth catching in CI regardless of what we decide here.

0xDEnYO:
Closing this — the enum approach does not work on-device, as documented in the comment above (ENUM_VALUE encodes an entry value in one byte, taken from the enum key, so no chain id above 255 is representable; `erc7730 calldata` drops the descriptor to `[]` while exiting 0).

We are reverting `destinationChainId` to `raw` on our side instead, which means **the existing sync PR #2901 will go green by itself** — its descriptor will regenerate as `raw` and match the `testsv2` fixture already on `master`. Verified locally against a real `master` checkout: both pinned runners 14/14 with the fixture untouched. So no fixture change is needed in this repo, and nothing here needs to be merged.

Leaving the thread intact because the measurements in it are the useful part, and opening a separate issue for the two upstream gaps rather than keeping them buried in a closed PR. Thanks for the patience on the churn.

Issue #2916, opened 2026-08-24
format: chainId has no defined fallback or pinned name source, and enum cannot substitute (ENUM_VALUE is one byte)


0xDEnYO:
Two gaps that make `format: chainId` unusable for a cross-chain descriptor, and one that makes `format: enum` an unsafe substitute. All measured against the two runners this repo pins in `.github/actions/run-*-tests/action.yml` (`sourcifyeth/clear-signing-test-runner@dae3cda`, `llbartekll/clear-signing@10605ba`) and against `erc7730` 1.0.10.

Context: LI.FI's `registry/lifi/calldata-LIFIDiamond.json` renders `_bridgeData.destinationChainId`. We switched it to `chainId`, could not make it work, tried `enum`, could not make that work either, and have reverted to `raw`. Writing up what we found rather than just reverting quietly, because we do not think this is specific to us.

## 1. `chainId` has no defined fallback, so implementations diverge

`specs/erc-7730.md:1411-1416` defines `chainId` as *"Value is converted to a Blockchain name using EIP-155 reference values"*. Unlike its neighbours it specifies no behaviour for a value it cannot resolve — compare `tokenAmount` (`:1356`, *"the wallet SHOULD display the raw value instead with an 'Unknown token' warning"*) and `nft` (`:1369`, *"fallback to a raw int token ID if not"*).

We probed 30 chain ids. **7 of 30 render identically across the two runners:** `100`, `8453`, `42161`, `42170`, `59144`, `534352`, `7777777`. A sample of the rest:

| chain id | Sourcify | Rust |
| --- | --- | --- |
| `1` | `Ethereum Mainnet` | `Ethereum` |
| `10` | `OP Mainnet` | `Optimism` |
| `56` | `BNB Smart Chain Mainnet` | `BNB Chain` |
| `137` | `Polygon Mainnet` | `Polygon` |
| `1868` | `Soneium` | `Chain 1868` |
| `1151111081099710` | `1151111081099710` | `Chain 1151111081099710` |

Note the last row: the two implementations disagree even on the *shape* of the unknown-id fallback (bare decimal vs `Chain <id>`). Both are defensible precisely because neither is specified.

[...]

PR #2917, opened 2026-08-24
ci: pin the Python tooling installed by CI

fbwoolf:
`erc7730` is installed unpinned in `validate_descriptors` (pull_request.yml), `format.yml` and `master.yml`, and `check-jsonschema` unpinned in `validate_schemas`. A release of either can turn every open pull request red with no commit in this repo. Reported by @0xDEnYO in #2916.

## Change

- Pin `erc7730==1.0.10` in `.github/requirements.txt`, which already existed carrying an unpinned `erc7730` and was referenced by nothing.
- Pin `check-jsonschema==0.38.0` in a new `.github/requirements-schema.txt`, kept separate so `validate_schemas` does not install the full linter.
- Install both with `pip install -r`.
- Add the `pip` ecosystem to dependabot on `/.github`. Its pip updater scans the directory, so both files are covered.

Both pins are the versions CI resolves to today, so this freezes current behaviour rather than changing it. Upgrades now arrive as dependabot pull requests that run the full suite before merge.

Note for #2916: the third `erc7730` site is `master.yml`, not `validate_attested_descriptors`, which is shell and jq and installs nothing.

<sub><i>Opened with Claude Code</i></sub>


fbwoolf:
Green here does not exercise the change: both install steps are conditional on changed descriptors or changed JSON, and this pull request has neither, so `validate descriptors` and `validate JSON schemas` short-circuit before installing.

Verified locally instead, on Python 3.12 to match the workflows:

| requirements file | resolves |
| --- | --- |
| `.github/requirements.txt` | `erc7730==1.0.10`, 46 packages |
| `.github/requirements-schema.txt` | `check-jsonschema==0.38.0` |

Note `erc7730` declares `requires_python >=3.12,<3.13`, so the `python-version: "3.12"` already set in all three workflows is load-bearing, not incidental.

kuzdogan (approved):

PR #2945, opened 2026-09-01
docs: add disclaimer on registry inclusion and attestation scope

fbwoolf:


hesterbruikman (approved):
Thanks for adding this!

PR #2965, opened 2026-09-09
ci: check that every function of a calldata descriptor has a test

### LINKED ISSUE #2964: CI does not check that every function in a descriptor has a test case
A calldata descriptor lists the functions it formats under `display.formats`, and its `testsv2` file holds the transactions the runners exercise. Nothing checks that the two agree. A descriptor can declare five functions while its test file only covers one, and CI stays green: the require-testsv2 job only checks that the file exists, and the runners only exercise the functions that a fixture calls. A format nobody tests is never verified against an implementation, so a wrong path or label in it ships unnoticed.

A full run over the registry today shows that 117 of the 190 calldata descriptors with a test file have at least one function without a test.

We want a CI check on pull requests that:

- computes the 4-byte selector of every key in `display.formats` of each affected calldata descriptor, with parameter names stripped as the ERC-7730 spec prescribes;
- reads the selector from the first 4 bytes of the calldata of every `rawTx` in the matching `testsv2` file;
- fails when a format has a selector that no test case contains, naming the missing functions.

Like the other validation jobs it should only run on the descriptors a pull request affects (changed descriptor, changed shared file it includes, or changed test file). EIP-712 descriptors have no selector and are out of scope. A descriptor with no test file at all is already reported by require-testsv2, so it should not be reported twice.

*Posted with Claude Code*

marcocastignoli:
Closes #2964.

## What

A new `check-test-coverage` job in `clear-signing-tests.yml`, next to `require-testsv2`, runs `.github/scripts/check-selector-coverage.js` on the affected descriptors that have a test file. For each one the script resolves `includes`, computes the 4-byte selector of every `display.formats` key, decodes every `rawTx` of the matching `testsv2` file and takes the first 4 bytes of its calldata. A format whose selector no test contains fails the job with an annotation on the descriptor that names the missing functions and their selectors, and the results comment lists the gaps next to the missing test files.

## How

- The job consumes the `test_matrix` output of `detect-testsv2`, so it only sees descriptors with a test file and does not repeat the changed-files step.
- Like the rest of the workflow it runs the script from the base branch, behind a `hashFiles` guard so this pull request passes with the job skipped. The base sparse checkout includes `package.json` and the lockfile, and `npm ci` runs inside it, so the base script gets the base dependencies.
- viem does the parsing on both sides: `parseAbiItem` plus `toFunctionSelector` for the format keys (parameter names stripped, `uint` expanded to `uint256`, tuples handled) and `parseTransaction` for the fixtures (legacy and typed envelopes, signed or unsigned). It is added as a devDependency.
- The script also fails on a test without `rawTx`, a test whose selector matches no format of the descriptor, a fixture whose `descriptor` field names another descriptor, a fixture that cannot be decoded, a format key that is not a function signature, and a descriptor that cannot be read or resolved. EIP-712 descriptors are skipped.
- The job uploads a JSON report as the `test-coverage` artifact; `clear-signing-tests-results.yml` renders it in the pull request comment.

## Verification

- All 605 calldata fixtures in the registry decode, every one of their selectors matches a format of its descriptor, and every fixture names its own descriptor.
- A full-registry run reports 117 of 190 tested calldata descriptors with at least one untested function. Those are existing gaps, and they only fail the job once someone touches the descriptor, same as require-testsv2.
- The job is skipped on this pull request, so the comment rendering was checked locally by running the workflow's inline script against mocked GitHub calls, a context and a coverage report.

## Docs

The "Pull Request content requirements" section of the README and `docs/REVIEWING.md` mention the new check.

### DIFF
--- a/.github/scripts/check-selector-coverage.js
+++ b/.github/scripts/check-selector-coverage.js
@@ -141,12 +141,19 @@ function checkDescriptor(descriptorAbs) {
 
   const covered = new Set();
   tests.forEach((test, i) => {
+    const name = test?.description ? JSON.stringify(test.description) : `#${i + 1}`;
+    let selector;
     try {
-      covered.add(testSelector(test));
+      selector = testSelector(test);
     } catch (error) {
-      const name = test?.description ? JSON.stringify(test.description) : `#${i + 1}`;
       errors.push(`Test ${name} in ${testFile} has a rawTx that cannot be decoded: ${reason(error)}`);
+      return;
     }
+    if (!selectors.has(selector)) {
+      errors.push(`Test ${name} in ${testFile} calls ${selector}, which no format of the descriptor matches`);
+      return;
+    }
+    covered.add(selector);
   });
 
   const uncovered = [...selectors].filter(([selector]) => !covered.has(selector));
--- a/.github/scripts/check-selector-coverage.js
+++ b/.github/scripts/check-selector-coverage.js
@@ -11,9 +11,10 @@
  * selector no test calls is an error. EIP-712 descriptors have no selector
  * and are skipped.
  *
- * Usage: node check-selector-coverage.js <descriptor.json|directory>...
+ * Usage: node check-selector-coverage.js [--report <file>] <descriptor.json|directory>...
  *   A directory is walked for calldata-*.json descriptors, skipping the
- *   tests/, testsv2/ and sigs/ folders.
+ *   tests/, testsv2/ and sigs/ folders. With --report, the errors are also
+ *   written to <file> as JSON, keyed by descriptor, for the results comment.
  *
  * Prints GitHub Actions annotations, and exits 1 when a function has no test,
  * a format key is not a function signature, or a test cannot be decoded.
@@ -21,6 +22,7 @@
 
 const fs = require('fs');
 const path = require('path');
+const { parseArgs } = require('util');
 const { parseAbiItem, parseTransaction, slice, toFunctionSelector } = require('viem');
 const { resolveDescriptor } = require('./resolve-erc7730-includes');
 
@@ -76,11 +78,7 @@ function testSelector(test) {
 
 // --- Check -------------------------------------------------------------------
 
-/**
- * The errors of one descriptor, or null when there is nothing to check: the
- * descriptor formats no function, or it has no test file (the require-testsv2
- * job reports it).
- */
+/** The errors of one descriptor, or null when it formats no function. */
 function checkDescriptor(descriptorAbs) {
   const descriptor = rel(descriptorAbs);
   const errors = [];
@@ -111,12 +109,7 @@ function checkDescriptor(descriptorAbs) {
   try {
     fixture = JSON.parse(fs.readFileSync(path.join(repoRoot, testFile), 'utf8'));
   } catch (error) {
-    if (error.code === 'ENOENT') {
-      // require-testsv2 reports the missing file; it is a different fix.
-      process.stderr.write(`warning: ${descriptor} has no test file ${testFile}\n`);
-      return errors.length > 0 ? errors : null;
-    }
-    errors.push(`Cannot read ${testFile}: ${error.message}`);
+    errors.push(`Cannot read ${testFile}: ${error.code === 'ENOENT' ? 'no such file' : error.message}`);
     return errors;
   }
 
@@ -167,15 +160,19 @@ function checkDescriptor(descriptorAbs) {
 }
 
 function main() {
-  const targets = process.argv.slice(2);
+  const { values, positionals: targets } = parseArgs({
+    options: { report: { type: 'string' } },
+    allowPositionals: true,
+  });
   if (targets.length === 0) {
-    process.stderr.write('Usage: node check-selector-coverage.js <descriptor.json|directory>...\n');
+    process.stderr.write('Usage: node check-selector-coverage.js [--report <file>] <descriptor.json|directory>...\n');
     process.exit(2);
   }
 
   const descriptors = [...new Set(targets.flatMap((t) => collectDescriptors(path.resolve(t), [])))].sort();
   let checked = 0;
   let failed = 0;
+  const report = {};
   for (const descriptorAbs of descriptors) {
     const descriptor = rel(descriptorAbs);
     if (!/^calldata-/.test(path.basename(descriptor))) continue;
@@ -190,7 +187,12 @@ function main() {
       continue;
     }
     failed++;
+    report[descriptor] = result;
     for (const message of result) console.log(`::error file=${descriptor},line=1::${message}`);
+  }
+  if (values.report) {
+    fs.mkdirSync(path.dirname(values.report), { recursive: true });
+    fs.writeFileSync(values.report, JSON.stringify(report, null, 2));
   }
 
   const summary = `${failed} of ${checked} descriptor(s) failed the check.`;
--- a/.github/scripts/check-selector-coverage.js
+++ b/.github/scripts/check-selector-coverage.js
@@ -78,8 +78,8 @@ function testSelector(test) {
 
 /**
  * The errors of one descriptor, or null when there is nothing to check: the
- * descriptor cannot be read (another job reports it), it formats no function,
- * or it has no test file (the require-testsv2 job reports it).
+ * descriptor formats no function, or it has no test file (the require-testsv2
+ * job reports it).
  */
 function checkDescriptor(descriptorAbs) {
   const descriptor = rel(descriptorAbs);
@@ -89,9 +89,7 @@ function checkDescriptor(descriptorAbs) {
   try {
     formats = resolveDescriptor(descriptorAbs).display?.formats ?? {};
   } catch (error) {
-    // Another job reports the malformed descriptor; there is nothing to cover.
-    process.stderr.write(`warning: cannot read ${descriptor}: ${error.message}\n`);
-    return null;
+    return [`Cannot read the descriptor: ${error.message}`];
   }
 
   const selectors = new Map(); // selector -> the keys that hash to it
@@ -173,7 +171,7 @@ function main() {
     for (const message of result) console.log(`::error file=${descriptor},line=1::${message}`);
   }
 
-  const summary = `${failed} of ${checked} descriptor(s) have a function without a test.`;
+  const summary = `${failed} of ${checked} descriptor(s) failed the check.`;
   console.log(summary);
   if (process.env.GITHUB_STEP_SUMMARY && failed > 0) {
     // GitHub shows 10 annotations per step, so give the total as well.
--- a/.github/scripts/check-selector-coverage.js
+++ b/.github/scripts/check-selector-coverage.js
@@ -65,9 +65,10 @@ function selectorOf(key) {
   return toFunctionSelector(parseAbiItem(`function ${key.trim()}`));
 }
 
-/** The selector called by a test case, or null when the test is not a calldata test. */
+/** The selector called by a test case. */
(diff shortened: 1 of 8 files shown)


.github/workflows/pull_request.yml
```diff
@@ -214,6 +214,67 @@ jobs:
...
+      - name: Resolve affected descriptors
+        timeout-minutes: 5
+        id: affected-descriptors
+        if: steps.changed-files.outputs.any_changed == 'true'
+        env:
+          CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
+          DELETED_FILES: ${{ steps.changed-files.outputs.deleted_files }}
+        run: |
+          # A changed shared file or testsv2 file affects the descriptors that
+          # include it or that it tests, so those are checked too.
```
manuelwedler: Following the `includes` chain makes a change to a shared file fail on every pre-existing gap of its includers, and the PR that changed the shared file cannot fix those gaps.

Measured on this branch with a full-registry run:

| changed shared file | affected calldata descriptors | fail the coverage check today |
|---|---|---|
| `ercs/calldata-erc4626-vaults.json` | 134 | 75 |
| `registry/kiln/common-KilnVaults.json` | 28 | 24 |
| `registry/safe/common-Safe.json` | 6 | 3 |

A PR may only touch one entity, so a PR that edits `ercs/calldata-erc4626-vaults.json` cannot add the missing Morpho, Kiln, etc. tests. The job stays red until every includer has tests, which blocks changes to the shared files.

The same applies to a PR that only edits a `common-*.json` file of an entity with many descriptors.

Suggestion: fail only for descriptors whose own file or own `testsv2` file changed in the PR. For descriptors that are affected only through the include chain, print a warning (`::warning`) instead of an error, or skip them. `list-affected-descriptors.js` does not tell the two apart, so the workflow would need to pass the directly changed files separately, or the script would need a flag.

README.md
```diff
@@ -195,7 +195,7 @@ Test files should be placed in a `testsv2/` folder within your entity directory
 
 ### Best practices
 
-1. **Include at least one test per function/message type** defined in your descriptor
+1. **Include at least one test per function/message type** defined in your descriptor. For calldata descriptors CI enforces this: it computes the 4-byte selector of every key in `display.formats` and compares it with the first 4 bytes of the calldata of each test case's `rawTx`. A function whose selector no test case contains fails the check.
```
manuelwedler: I think this addition is not needed and misplaced.
```suggestion
1. **Include at least one test per function/message type** defined in your descriptor
```

Rather tell about it in the `Pull Request content requirements` section.

.github/workflows/pull_request.yml
```diff
@@ -214,6 +214,67 @@ jobs:
         # sync-indexes workflow — by then the conflict is already on master.
         run: node tools/scripts/generate-index.js --validate
 
+  validate_test_coverage:
```
manuelwedler: Would it be more reasonable to move this job to `clear-signing-tests.yml`?

- The `detect-testsv2` job there already has this exact changed-files filter and the affected-descriptors step, and it exports `affected_descriptors` and `test_matrix` as outputs. This job duplicates that step.
- The check is a sibling of `require-testsv2`: that job checks that each affected descriptor has a test file, this one checks that the test file covers every function. It could consume `test_matrix`, which only lists descriptors with a test file, so the "no test file" branch of the script goes away.
- The results workflow already reports the missing test files in the PR comment. Coverage gaps fit the same comment.

Two things to keep in mind:

- That workflow runs the scripts from the PR base (`base-scripts` checkout), so the step needs the same `hashFiles` guard as the recommendations step, so that this PR passes.
- The detect job has no Node setup and no `npm ci`. The check needs viem, so it needs its own job after `detect-testsv2`, or the setup steps added to the detect job.

The workflow only runs when a JSON file under `registry/` or `ercs/` changes, which is the only case where this check has work to do.

.github/scripts/check-selector-coverage.js
```diff
@@ -0,0 +1,189 @@
...
+ * signature with parameter names, so the ABI parser strips the names and
+ * expands shorthands like "uint" before hashing.
+ */
+function selectorOf(key) {
+  return toFunctionSelector(parseAbiItem(`function ${key.trim()}`));
+}
+
+/** The selector called by a test case, or null when the test is not a calldata test. */
+function testSelector(test) {
+  if (typeof test.rawTx !== 'string') return null;
```
manuelwedler: This should fail rather than return null. The script only reaches this function for `calldata-*` descriptors, so a test without a string `rawTx` is either malformed or an EIP-712 test in the wrong file. Today such a test is silently ignored and the descriptor is judged on its other tests only.

The test schema does not catch every case: `tests` items are a `oneOf` of calldata test and EIP-712 test, so a test with `data` and no `rawTx` in a calldata fixture passes the schema check.

The loop in `checkDescriptor` already has an error path for a `rawTx` that cannot be decoded, so a throw here lands there and names the test:

```suggestion
  if (typeof test.rawTx !== 'string') throw new Error('test has no rawTx');
```

The `if (selector)` guard in the loop can go then, and `testSelector` always returns a string.

.github/scripts/check-selector-coverage.js
```diff
@@ -0,0 +1,189 @@
...
+  const descriptor = rel(descriptorAbs);
+  const errors = [];
+
+  let formats;
+  try {
+    formats = resolveDescriptor(descriptorAbs).display?.formats ?? {};
+  } catch (error) {
+    // Another job reports the malformed descriptor; there is nothing to cover.
+    process.stderr.write(`warning: cannot read ${descriptor}: ${error.message}\n`);
+    return null;
```
manuelwedler: This can fail too. A descriptor that cannot be read or resolved is a real problem for this check, and the job should not report it as "nothing to check" and pass. Other jobs report it as well, but the check does not need to depend on them. Same as for an unreadable test file a few lines below, which is already an error.

```suggestion
    return [`Cannot read the descriptor: ${error.message}`];
```

.github/scripts/check-selector-coverage.js
```diff
@@ -0,0 +1,189 @@
...
+      process.stderr.write(`warning: ${descriptor} has no test file ${testFile}\n`);
+      return errors.length > 0 ? errors : null;
+    }
+    errors.push(`Cannot read ${testFile}: ${error.message}`);
+    return errors;
+  }
+  if (!Array.isArray(tests)) {
+    errors.push(`${testFile} has no "tests" array`);
+    return errors;
+  }
```
manuelwedler, line 133: The fixture is paired with the descriptor by file name only. Every fixture also names the descriptor it tests in its `descriptor` field (`"descriptor": "../calldata-AggregationRouterV3.json"`), and the script never reads it.

Please add a check that the `descriptor` field, resolved from the fixture's directory, points at the descriptor under test, and report a mismatch as an error. It catches a fixture that was copied from another descriptor without updating the field: the script would otherwise count the shared functions as covered while the fixture says it tests the other contract.

All fixtures match today, so this only guards new contributions.

.github/scripts/check-selector-coverage.js
```diff
@@ -0,0 +1,189 @@
...
+  const covered = new Set();
+  tests.forEach((test, i) => {
+    try {
+      const selector = testSelector(test);
+      if (selector) covered.add(selector);
+    } catch (error) {
+      const name = test?.description ? JSON.stringify(test.description) : `#${i + 1}`;
+      errors.push(`Test ${name} in ${testFile} has a rawTx that cannot be decoded: ${reason(error)}`);
+    }
+  });
```
manuelwedler, line 150: The check only goes in one direction. A test whose selector matches no format of the descriptor is accepted without a word.

Please also report the reverse: a test whose selector is not in `selectors` is an error that names the test and the selector. Such a test calls a function the descriptor does not format, so it tests nothing of the descriptor, and it usually means the fixture holds a transaction of the wrong contract or the format key has a typo.

The PR description says all 605 fixtures match a format of their descriptor, so this passes today and only guards new contributions. The `selectors` map is already at hand in this loop.

manuelwedler (changes requested):

.github/scripts/check-selector-coverage.js
marcocastignoli: Done in a33e901: a test without a string `rawTx` now throws, which lands in the existing per-test error path and names the test. The `if (selector)` guard is gone.

*Posted with Claude Code*

.github/scripts/check-selector-coverage.js
marcocastignoli: Done in 7bbac88: a descriptor that cannot be read or resolved returns the error line you suggested instead of a skip.

*Posted with Claude Code*

.github/scripts/check-selector-coverage.js
marcocastignoli: Done in 0a89cbd: the script resolves the fixture's `descriptor` field from the fixture's directory and reports a mismatch, or a missing field, as an error on the descriptor. The coverage check stops there for that descriptor.

*Posted with Claude Code*

.github/scripts/check-selector-coverage.js
marcocastignoli: Done in 29a8f97: a test whose selector is not in the descriptor's formats is an error naming the test and the selector, and it does not count as coverage.

*Posted with Claude Code*

README.md
marcocastignoli: Done in 98c2588: the best-practices line is back to its original text, and the rule is now part of the existing test-file bullet under "Pull Request content requirements".

*Posted with Claude Code*

.github/workflows/pull_request.yml
marcocastignoli: Done in 625b0fd: the job is now `check-test-coverage` in `clear-signing-tests.yml`, after `detect-testsv2` and next to `require-testsv2`. It consumes `test_matrix`, so the "no test file" branch of the script is gone. It runs the base script behind the same `hashFiles` guard as the recommendations step, with `npm ci` run inside the base checkout so the script gets the base dependencies. It uploads a `test-coverage` artifact that the results workflow renders in the pull request comment, right after the missing test files block. The job is skipped on this pull request, so the comment rendering was checked locally by running the workflow's inline script against mocks.

*Posted with Claude Code*

marcocastignoli: Correction to the above: you asked for "the same `hashFiles` guard as the recommendations step, so that this PR passes". I had put the guard on the job condition, where `hashFiles` is not available, so the push of 625b0fd made the workflow file invalid and the run failed before it started. Fixed in 985fff0: the guard is now on the check and upload steps, like the recommendations step, and the job condition only requires an affected descriptor with a test file.

*Posted with Claude Code*

.github/workflows/clear-signing-tests.yml
```diff
@@ -391,6 +391,58 @@ jobs:
           entity: ${{ matrix.test.entity }}
           descriptor-name: ${{ matrix.test.descriptor_name }}
 
+  check-test-coverage:
```
manuelwedler: This job never ran in CI. The tests workflow only triggers on a change under `registry/` or `ercs/`, and the check step is skipped by the `hashFiles` guard until the script is on master. The only run on this branch is the failed one from 625b0fd. Locally, the workflow files pass actionlint and the sparse checkout of the root package files works in cone mode, but the job itself, the artifact hand-off and the coverage block in the results comment are unverified.

Please verify the full path once on your fork before merge: merge this branch into the fork's master, open a PR on the fork with a registry change, and check that `check-test-coverage` runs, that the artifact is uploaded, and that the results comment shows the coverage gaps. A link to that run in the PR description is enough.

manuelwedler (approved):
lgtm

PR #2853, opened 2026-08-05
fix(permit): show unlimited approvals as "Unlimited"

fbwoolf:
Adds a `threshold` and `message: "Unlimited"` to the `value` field in the shared ERC-2612 permit descriptor (`ercs/eip712-erc2612-permit.json`), which every permit descriptor includes. Previously an unlimited approval rendered as a raw 78-digit number with no clear cue to the user.

The threshold uses `0x8000...0000` (2^255) to match the repo's existing convention (e.g. `registry/lido/calldata-stETH.json`).

## Test coverage

Every existing permit fixture used an ordinary finite amount. The largest was `2.5e23`, about 10^53 below the threshold, so no test reached the new branch. CI proved that nothing broke. It did not prove that "Unlimited" ever rendered.

`registry/permit/testsv2/eip712-permit-arbitrum-arb.tests.json` now carries a second case, with `value` set to `2^256 - 1`. Both runners render `Unlimited ARB`.

Negative control: revert `ercs/eip712-erc2612-permit.json` to master, and the new case fails with

```
expected "Unlimited ARB"
got      "115792089237316195423570985008687907853269984665640564039457.584007913129639935 ARB"
```

That string is the defect the audit reported.

## Closing list

This pull request closes 43 issues. Each keyword is repeated, because GitHub applies a closing keyword only to the reference that follows it directly.

**25 issues whose only finding is the missing `threshold`:**

Closes #2727
Closes #2731
Closes #2733
Closes #2736
Closes #2737
Closes #2743
Closes #2752
Closes #2754
Closes #2755
Closes #2758
Closes #2761
Closes #2767
Closes #2770
Closes #2774
Closes #2775
Closes #2776
Closes #2779
Closes #2781
Closes #2782
Closes #2785
Closes #2787
Closes #2789
Closes #2792
Closes #2793
Closes #2799

**18 issues that carry a second finding, an EIP-712 domain mismatch. #2856 (correct EIP-712 domains to match on-chain tokens) fixed that half on 2026-08-11, so this pull request clears the remainder:**

Closes #2728
Closes #2729
Closes #2734
Closes #2735
Closes #2738
Closes #2741
Closes #2744
Closes #2748
Closes #2763
Closes #2768
Closes #2771
Closes #2772
Closes #2773
Closes #2777
Closes #2778
Closes #2783
Closes #2791
Closes #2797

[...]

PR #2967, opened 2026-09-09
fix(swell): show unlimited allowances as "Unlimited"

### LINKED ISSUE #2815: Cyfrin Audit: [Swell - calldata-swell]
**Descriptor:** [registry/swell/calldata-swell.json](https://github.com/ethereum/clear-signing-erc7730-registry/blob/a2b33ffed92aab99cb9da2704610469a2588b520/registry/swell/calldata-swell.json)

**ERC-8176 descriptor hash:** `0x9aa2999776fbe9ed2f813ec603358c4781ccccb88e91ecd14acb26363971369c`

**1: `approve` / `increaseAllowance` have no `threshold`; an unlimited allowance renders as a raw number, not "Unlimited"**
**Severity:** Medium

**Description:** Unlimited token approvals are the highest-persistence permission a user grants, and dApps request them routinely. Without a `threshold`, the signer cannot recognize the unlimited case at a glance; it looks like an oddly large finite amount. `increaseAllowance` has the same gap.

**Recommended mitigation:** Change the affected descriptor entries as follows:

```diff
// registry/swell/calldata-swell.json:78
+// add to both amount params:
+"params": {
+"token": "$.metadata.constants.swellToken", // or "tokenPath": "@.to"
+"threshold": "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
+"message": "Unlimited"
+}
```

venugopalanvip:
Closes #2815 — Cyfrin audit, Medium.

`approve` and `increaseAllowance` on `calldata-swell.json` had no `threshold`, so `type(uint256).max` rendered as a 78-digit integer instead of `Unlimited`. An unlimited allowance is the highest-persistence permission a user can grant, and it should be recognisable at a glance rather than looking like an oddly large finite amount.

## The change

Adds `threshold` and `message` to the amount field of both formats. Two entries touched, nothing else.

## A deliberate deviation from the suggested diff

The issue's recommended mitigation uses `0xffff…ffff` (`type(uint256).max`). I used **`0x8000…0000` (2^255)** instead, for two reasons:

1. **It is the registry's established convention.** `ercs/eip712-erc2612-permit.json` — the shared file that #2853 used to clear 74 findings of this exact kind — uses 2^255, as do 15 other descriptors. Only 2 use a max-uint256-style value.
2. **It catches more of the real cases.** A threshold is a floor, not an equality check. At 2^255 an allowance that is effectively unlimited but not exactly `type(uint256).max` still renders as `Unlimited`; at max-uint256 it would display as a 78-digit number and the finding would remain open in practice.

Happy to switch to the literal value from the issue if reviewers prefer the audit text verbatim.

## `decreaseAllowance` deliberately left alone

It has the same missing `threshold`, but the audit did not flag it and I think that is right: an unlimited *decrease* clears an allowance rather than granting one, so `Unlimited` there would describe a safe action in the language used for a dangerous one.

## Test

Added a reference test built from a real mainnet unlimited approve — [`0x1b3e99df…`](https://etherscan.io/tx/0x1b3e99df256414854cddd9ab688554066e1c66e0494c50a6c172ad99031e5de5), an exact `type(uint256).max` approval:

| Test | Amount renders as |
|---|---|
| Approve spender (existing) | `0.25 rswETH` |
| Transfer Tokens (existing) | `0.8856 rswETH` |
| **Approve spender unlimited (new)** | **`Unlimited rswETH`** |

The two existing cases are finite and unaffected, so they also serve as a regression check that ordinary amounts still render normally.

## Validation

```
erc7730 lint registry/swell/calldata-swell.json   -> 0 errors
erc7730 format registry/swell/calldata-swell.json -> no issue found
```

Both files validate against `specs/erc7730-v2.schema.json` and `specs/erc7730-tests-v2.schema.json`. The formatter expanded the two edited entries to multiple lines — that is `erc7730 format`'s own output for the longer params, not a hand reformat.

### DIFF
--- a/registry/swell/calldata-swell.json
+++ b/registry/swell/calldata-swell.json
@@ -23,7 +23,16 @@
         "intent": "Approve spender",
         "fields": [
           { "label": "Spender", "format": "addressName", "params": { "types": ["wallet", "eoa", "contract"] }, "path": "#.spender" },
-          { "label": "Amount", "format": "tokenAmount", "path": "amount", "params": { "token": "$.metadata.constants.swellToken" } }
+          {
+            "label": "Amount",
+            "format": "tokenAmount",
+            "path": "amount",
+            "params": {
+              "token": "$.metadata.constants.swellToken",
+              "threshold": "0x8000000000000000000000000000000000000000000000000000000000000000",
+              "message": "Unlimited"
+            }
+          }
         ]
       },
       "batchAddToWhitelist(address[] _addresses)": {
@@ -75,7 +84,16 @@
         "intent": "Increase allowance",
         "fields": [
           { "label": "Spender", "format": "addressName", "params": { "types": ["wallet", "eoa", "contract"] }, "path": "#.spender" },
-          { "label": "Added Value", "format": "tokenAmount", "path": "#.addedValue", "params": { "tokenPath": "@.to" } }
+          {
+            "label": "Added Value",
+            "format": "tokenAmount",
+            "path": "#.addedValue",
+            "params": {
+              "tokenPath": "@.to",
+              "threshold": "0x8000000000000000000000000000000000000000000000000000000000000000",
+              "message": "Unlimited"
+            }
+          }
         ]
       },
       "removeFromWhitelist(address _address)": {
--- a/registry/swell/testsv2/calldata-swell.tests.json
+++ b/registry/swell/testsv2/calldata-swell.tests.json
@@ -22,6 +22,16 @@
         "owner": "Swell",
         "fields": [{ "label": "To", "value": "0x7e702F7A8299b51D3Eb0A3E8107Ce64f7D726966" }, { "label": "Amount", "value": "0.8856 rswETH" }]
       }
+    },
+    {
+      "description": "Approve spender unlimited - chain 1",
+      "rawTx": "0x02f86d016a8424eec95a84299c0bb483013c3894fae103dc9cf190ed75350761e95403b7b8afa6c080b844095ea7b300000000000000000000000040aa958dd87fc8305b97f2ba922cddca374bcd7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffc0",
+      "txHash": "0x1b3e99df256414854cddd9ab688554066e1c66e0494c50a6c172ad99031e5de5",
+      "expected": {
+        "intent": "Approve spender",
+        "owner": "Swell",
+        "fields": [{ "label": "Spender", "value": "0x40aA958dd87FC8305b97f2BA922CDdCa374bcD7f" }, { "label": "Amount", "value": "Unlimited rswETH" }]
+      }
     }
   ]
 }


registry/swell/calldata-swell.json
```diff
@@ -75,7 +84,16 @@
...
         "fields": [
           { "label": "Spender", "format": "addressName", "params": { "types": ["wallet", "eoa", "contract"] }, "path": "#.spender" },
-          { "label": "Added Value", "format": "tokenAmount", "path": "#.addedValue", "params": { "tokenPath": "@.to" } }
+          {
+            "label": "Added Value",
+            "format": "tokenAmount",
+            "path": "#.addedValue",
+            "params": {
+              "tokenPath": "@.to",
+              "threshold": "0x8000000000000000000000000000000000000000000000000000000000000000",
```
marcocastignoli, line 86: This half of the finding has no test yet. When you add the `increaseAllowance` fixture for coverage, could it use an unlimited `addedValue`? A finite value would pass the coverage check but leave this threshold untested.

marcocastignoli (changes requested):
Requesting changes only because of the test coverage rule that landed after this branch was cut. The threshold change itself looks right.

What I checked:
- The new fixture's `rawTx` has the same `to` and calldata as mainnet transaction `0x1b3e99df…`, an exact `type(uint256).max` approval, and both runners render it as `Unlimited rswETH`.
- Using 2^255 instead of the audit's max-uint256 value is fine. It matches `ercs/eip712-erc2612-permit.json` and also catches allowances just below the maximum.
- Leaving `decreaseAllowance` alone is the right call, for the reason you gave.

**Blocking: test coverage.** The branch is 27 commits behind master, and its last CI run (2026-09-09) predates #2965. I ran master's `.github/scripts/check-selector-coverage.js` against this head, and it fails with 13 of 15 functions untested, the same list you posted. Please merge master and add the fixtures.

On your three questions, option 2 looks right to me:
- `allowance`: drop it from `display.formats`. It is a view function, so no signer ever sees it, and this PR changes the descriptor hash anyway.
- `burn` and `withdrawERC20`: give both an `intent` here, rather than writing fixtures around a missing one.
- `batchAddToWhitelist` / `batchRemoveFromWhitelist`: existing fixtures expect one field per array element, each with the same label. The "Merge artworks" case in `registry/feral-file/testsv2/calldata-feralfile-exhibition-v4_3-0.tests.json` shows the shape.

_Posted with Claude Code_

venugopalanvip:
Thanks — that is a genuinely useful review, and the three answers are what I needed. Taken option 2, in `5eeb6da`, on top of a merge of master.

**`allowance` dropped from `display.formats`.** Agreed on the reasoning: it is a view function, so no signer ever authorises it, and a fixture would have modelled a transaction nobody sends.

**`burn` and `withdrawERC20` given intents** — `Burn tokens` and `Withdraw ERC-20 token`. They were the only two formats in the file without one; every format now has an intent.

**Thirteen fixtures** for the twelve remaining functions. The extra one is a second `increaseAllowance` at `type(uint256).max`, so the threshold this PR adds is exercised rather than only declared:

```
Increase allowance unlimited     Added Value = Unlimited rswETH
```

The two array functions follow the shape you pointed at — one field entry per element, repeating the label:

```
Add several to whitelist   Addresses = 0x7e70…6966
                           Addresses = 0x40aA…cD7f
```

**Coverage now passes:** `13 of 15 function(s) have no test` becomes `0 of 14`.

Two notes on what I did **not** touch, so they are on the record rather than silently left:

- `depositViaDepositManager` renders `#._amount` with the native `amount` format, so its fixture expects `2 ETH` for what is an rswETH-denominated argument. That looks wrong to me in the same way the Lombard fee did, but it is a separate finding from this PR's `threshold` change and I did not want to widen the diff while you are unblocking it. Happy to open it separately.
- `decreaseAllowance` still has no `threshold`, deliberately, for the reason in the original description — lowering an allowance to a huge number is not an unlimited approval.

The rest of the branch is unchanged: same two `threshold` additions, same rationale for 2^255 over max-uint256.

marcocastignoli (approved):
Approving. 5eeb6da settles everything from my previous review:

- `Check test coverage` passes: each of the 14 remaining formats has a test, and `allowance` is gone.
- The new "Increase allowance unlimited" case renders `Unlimited rswETH`, so the second half of the finding is tested too.
- `burn` and `withdrawERC20` have intents, and the two whitelist batch cases follow the repeated-label shape.

I also checked the formats that still render raw integers (`Min Rsw ETH` in `depositViaDepositManager` and the three `reprice` fields). They are fine to leave: in `RswETH.sol` the first reverts unless the caller is the DepositManager contract, and `reprice` needs the REPRICER role, so no user signs either.

_Posted with Claude Code_

venugopalanvip:
A correction to my previous comment, since I put a claim on the record that turns out to be wrong.

I said `depositViaDepositManager` renders `#._amount` with the native `amount` format and that this "looks wrong in the same way the Lombard fee did". I had not checked the contract. `IrswETH.sol`, in the Sourcify match of the implementation behind the proxy at `0x4796d939…2362`, documents it:

```solidity
/**
 * @dev Called when a user deposits a LST into the deposit manager.
 * @param _amount The amount of ETH to mint rswETH for.
 * @param _to The address to mint rswETH for.
 * @param _minRsweTH The minimum amount of rswETH to mint.
 */
function depositViaDepositManager(uint256 _amount, address _to, uint256 _minRsweTH) external;
```

`_amount` is ETH-denominated, so `"format": "amount"` is correct and the fixture's `2 ETH` is right. My apologies for the noise.

The field that *is* wrong is the next one down: `_minRswETH` is "the minimum amount of rswETH to mint" and renders as `raw`, so the fixture I added expects `1900000000000000000` where it should read `1.9 rswETH`. That is the same class as the `assets` fields in Morpho Blue (#3009), and the swell descriptor already has `$.metadata.constants.swellToken` to point a `tokenAmount` at.

I have deliberately **not** folded it into this branch — you are blocking only on coverage and the branch is green now, so widening it again would not help. Happy to send it as its own PR once this merges, or to add it here if you would rather it went in together.

PR #2968, opened 2026-09-09
fix(opensea): make the Wyvern order screen state what is authorized

### LINKED ISSUE #2665: Cyfrin Audit: [Opensea - eip712-opensea]
**Descriptor:** [registry/opensea/eip712-opensea.json](https://github.com/ethereum/clear-signing-erc7730-registry/blob/08b9d5177d08109c6dbacde2c00220ef6129ba64/registry/opensea/eip712-opensea.json)

**ERC-8176 descriptor hash:** `0x9e8dda710d9f4dc2decdf049a4bb79cd02268da53ba14d2f0b329c377b8dc959`

**1: The Wyvern Order display does not identify the action being authorized**

**Severity:** Critical

**Description:** A signer cannot tell whether they are buying or selling, which asset moves, whether the order can delegatecall, whether the executed calldata can be rewritten, or what currency and auction terms apply. This is not a display-polish issue; the screen can show a confident listing label while hiding the fields that define the order's effect.

**Recommended mitigation:** Replace the current three-field screen with a constrained order display. At minimum, use a neutral intent and render the side, maker, taker, asset target, execution mode, calldata or decoded transfer, replacement pattern, payment token, price, sale kind,  listing time, and expiration time. If the descriptor cannot constrain or decode `target` + `calldata` + `replacementPattern` well enough for a signer to understand the asset transfer, then consider removing the `Order` format rather than showing a partial clear-signing screen.

```diff
// Apply at the affected descriptor field; do not remove surrounding object keys.
// registry/opensea/eip712-opensea.json:9
+ "metadata": {
+ "owner": "Wyvern Exchange Contract",
+ "enums": {
+ "side": { "0": "Buy", "1": "Sell" },
+ "saleKind": { "0": "Fixed Price", "1": "Dutch Auction" },
+ "howToCall": { "0": "Call", "1": "DelegateCall" }
+ }
+ },

### PR DESCRIPTION
**Pull request by venugopalanvip**
Closes #2665 — Cyfrin audit, and the audit's only **Critical**.

## The problem

The screen showed three fields under a fixed intent of `OpenSea Listing`:

```
OpenSea Listing
Contract address   0x7f268357A8c2552623316e2562D90e642bB538E5
Price              7500000000000000000
Offer expiration   1775174400
```

Everything that defines the order's effect was `visible: never` — `side`, `saleKind`, `howToCall`, `target`, `calldata`, `replacementPattern`, `paymentToken`, `maker`, `taker`.

The consequences are what make this Critical rather than cosmetic:

- **A buy renders identically to a sell.** `side` was hidden, but the intent asserted "Listing". The screen could state the opposite of what was being signed.
- **A `DelegateCall` order renders identically to a `Call` order.** `howToCall` was hidden.
- **The currency was invisible.** `basePrice` was `raw`, so 7.5 WETH and 7.5 ETH looked the same.
- **The asset was invisible.** `target` and `calldata` — which actually move the NFT — were both hidden.

## Why fix rather than remove

The issue offers removing the `Order` format as an alternative if the transfer cannot be decoded well enough. I checked whether Wyvern is still live before choosing:

```
atomicMatch_   2025-12-19
atomicMatch_   2025-10-11
atomicMatch_   2025-09-23
```

Residual but real usage. Removing the descriptor would push those signers back to full blind signing, which is strictly worse than an honest partial screen. So this fixes the display instead — with the important change that the screen no longer *claims* to know what the order is.

## The change

Intent becomes the neutral **"Sign Wyvern order"** rather than asserting a listing, and 11 decision-relevant fields become visible:

| Field | Format | Was |
|---|---|---|
| Order side | `enum` → Buy / Sell | hidden |
| Sale kind | `enum` → Fixed price / Dutch auction | hidden |
| Price | `tokenAmount` via `tokenPath: paymentToken` | `raw` integer |
| Maker / Taker | `addressName` | hidden |
| Asset contract | `addressName` | hidden |
| Execution mode | `enum` → Call / **DelegateCall** | hidden |
| Executed calldata | `raw`, visible | hidden |
| Rewritable bytes | `raw`, visible | hidden |
| Listing time / Expires | `date` | `raw` timestamp / hidden |

Two conventions borrowed from existing descriptors rather than invented: `metadata.enums` with `$ref` follows `aave/calldata-lpv2.json`, and `tokenAmount` + `tokenPath` pointing at another message field follows `1inch/eip712-1inch-limit-order.json`. Price now carries its own ticker, so the payment token is rendered without spending a separate screen on it.

`calldata` and `replacementPattern` are shown as `raw`. A descriptor cannot decode arbitrary nested calldata, but a signer seeing a non-empty replacement pattern at least knows the executed payload can be rewritten — which is invisible today.

## Left hidden, deliberately

Fees, `salt`, `nonce`, `staticTarget`, `staticExtradata`, and `extra`. `extra` is the Dutch-auction price delta and is arguably decision-relevant when `saleKind` is `Dutch auction`; I left it out to keep this change to the audit's stated list and to hold the screen count down. Happy to add it if reviewers want the auction terms complete.

## Tests

The existing case is re-expected against the new display. A second case is added for the scenario the finding describes — **a buy order that delegatecalls, with a non-empty replacement pattern**:

| | Case 1 | Case 2 |
|---|---|---|
| Order side | Sell | **Buy** |
| Sale kind | Fixed price | Dutch auction |
| Execution mode | Call | **DelegateCall** |
| Rewritable bytes | `0x` | `0xffff…ffff` |
| Price | 7.5 WETH | 1.25 WETH |

Under the previous descriptor **both of these rendered the same three lines** and both said "OpenSea Listing". That is the regression this is meant to prevent.

## Validation

```
erc7730 lint registry/opensea/eip712-opensea.json   -> 0 errors
erc7730 format registry/opensea/eip712-opensea.json -> no issue found
```

Both files validate against `specs/erc7730-v2.schema.json` and `specs/erc7730-tests-v2.schema.json`. The one remaining lint warning (`owner` exceeds 22 characters) pre-dates this change.

### DIFF
--- a/registry/opensea/eip712-opensea.json
+++ b/registry/opensea/eip712-opensea.json
@@ -6,35 +6,91 @@
       "domain": { "name": "Wyvern Exchange Contract", "version": "2.3" }
     }
   },
-  "metadata": { "owner": "Wyvern Exchange Contract" },
+  "metadata": {
+    "owner": "Wyvern Exchange Contract",
+    "enums": {
+      "side": { "0": "Buy", "1": "Sell" },
+      "saleKind": { "0": "Fixed price", "1": "Dutch auction" },
+      "howToCall": { "0": "Call", "1": "DelegateCall" }
+    }
+  },
   "display": {
     "formats": {
       "Order(address exchange,address maker,address taker,uint256 makerRelayerFee,uint256 takerRelayerFee,uint256 makerProtocolFee,uint256 takerProtocolFee,address feeRecipient,uint8 feeMethod,uint8 side,uint8 saleKind,address target,uint8 howToCall,bytes calldata,bytes replacementPattern,address staticTarget,bytes staticExtradata,address paymentToken,uint256 basePrice,uint256 extra,uint256 listingTime,uint256 expirationTime,uint256 salt,uint256 nonce)": {
-        "intent": "OpenSea Listing",
+        "intent": "Sign Wyvern order",
         "fields": [
-          { "path": "exchange", "label": "Contract address", "format": "raw" },
-          { "path": "basePrice", "label": "Price", "format": "raw" },
-          { "path": "expirationTime", "label": "Offer expiration", "format": "raw" },
-          { "label": "Side", "path": "side", "visible": "never" },
-          { "label": "Extra", "path": "extra", "visible": "never" },
-          { "label": "Sale Kind", "path": "saleKind", "visible": "never" },
-          { "label": "Static Target", "path": "staticTarget", "visible": "never" },
-          { "label": "Target", "path": "target", "visible": "never" },
-          { "label": "How To Call", "path": "howToCall", "visible": "never" },
-          { "label": "Maker Relayer Fee", "path": "makerRelayerFee", "visible": "never" },
-          { "label": "Fee Method", "path": "feeMethod", "visible": "never" },
-          { "label": "Replacement Pattern", "path": "replacementPattern", "visible": "never" },
-          { "label": "Taker Relayer Fee", "path": "takerRelayerFee", "visible": "never" },
-          { "label": "Taker", "path": "taker", "visible": "never" },
-          { "label": "Listing Time", "path": "listingTime", "visible": "never" },
-          { "label": "Static Extradata", "path": "staticExtradata", "visible": "never" },
-          { "label": "Maker Protocol Fee", "path": "makerProtocolFee", "visible": "never" },
-          { "label": "Taker Protocol Fee", "path": "takerProtocolFee", "visible": "never" },
-          { "label": "Payment Token", "path": "paymentToken", "visible": "never" },
-          { "label": "Calldata", "path": "calldata", "visible": "never" },
-          { "label": "Salt", "path": "salt", "visible": "never" },
-          { "label": "Fee Recipient", "path": "feeRecipient", "visible": "never" },
-          { "label": "Maker", "path": "maker", "visible": "never" }
+          {
+            "path": "side",
+            "label": "Order side",
+            "format": "enum",
+            "params": { "$ref": "$.metadata.enums.side" },
+            "visible": "always"
+          },
+          {
+            "path": "saleKind",
+            "label": "Sale kind",
+            "format": "enum",
+            "params": { "$ref": "$.metadata.enums.saleKind" },
+            "visible": "always"
+          },
+          {
+            "path": "basePrice",
+            "label": "Price",
+            "format": "tokenAmount",
+            "params": { "tokenPath": "paymentToken" },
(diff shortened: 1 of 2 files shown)


kuzdogan:
Thanks for this one. The enum maps, the domain and the `Order` type string all check out against the verified source, and the new screen is a big step up. I pushed `ee6d61b` and `4d324ff` on top of your branch with two things a review turned up, three fixtures, and a format change on the inner call.

**1. An ETH-priced order rendered as an unknown token.** `atomicMatch` treats a zero `paymentToken` as native ETH (it requires `msg.value` and pays the maker with `transfer`), and that is the common Wyvern order. `basePrice` had `tokenPath: paymentToken` with no `nativeCurrencyAddress`, so the runner rendered a 7.5 ETH listing as `7500000000000000000`, the same blindness #2665 filed. Both fixtures used WETH, so the tests never reached that branch. The commit adds the `addressAsNull` constant that `safe/common-Safe.json` uses and references it from the field. The new ETH-priced fixture renders `7.5 ETH`; against your head it fails with `expected "7.5 ETH", got "7500000000000000000"`.

**2. `extra` was hidden.** `calculateFinalPrice` moves the price by `extra` over a Dutch auction, down for a sell and up for a buy, so it is a value-bearing amount. It now shows as `Price change` in the payment token, with the same native-currency handling. A fixed-price order shows `0 WETH`, which is true and cheap.

**3. `calldata` is now `format: "calldata"` with `calleePath: "target"`.** It is an embedded call to the collection, so the wallet should try to render it with the collection's descriptor. No collection is bound in the registry today, so both runners fall back to the raw hex and the fixtures pin that; the day a collection descriptor lands, the transfer decodes here for free. The label is the neutral `Asset call`, and `Rewritable bytes` stays beside it, because a sell order leaves the buyer word zeroed and masks it in `replacementPattern`, so a decoded `to` on the signed bytes is a placeholder.

**Fixtures.** Three added: the ETH-priced order, a Dutch auction with a non-zero `extra`, and an order with a zero `taker`, the normal "anyone can fill" value. All five pass on the Sourcify runner at the commit CI pins. The zero taker renders as the zero address; v2 cannot render one value two ways, so that is pinned rather than hidden.

**Legacy fixture.** `tests/eip712-opensea.tests.json` still expected the old three-line screen. The v1 runner is disabled in CI and its device-wrapped texts cannot be regenerated, so I removed it; `testsv2/` covers the format.

Two things I left alone, for the auditors rather than for this PR: `makerRelayerFee` and `feeRecipient` stay hidden although they reduce what the maker receives, and the asset transfer is still raw calldata. The audit's own fallback is to remove the format if the transfer cannot be made readable. Usage of the contract is near zero, so that stays a live option.

kuzdogan (approved):

PR #2976, opened 2026-09-10
ci: show the rendered fields and the declared format of each passing case in the results comment

### LINKED ISSUE #2966: clearsig-ci: include rendered output and field formats of passing tests in the PR results comment
## Problem

Reviewing a descriptor PR today means reading two files side by side: the descriptor JSON (which declares each field's `format`, its `params` and the `intent` / `interpolatedIntent` template) and the `testsv2/*.tests.json` file (which holds the raw transaction and the expected rendered output). The test files are long, the `rawTx` blobs push the interesting `expected` blocks far apart, and the descriptor's field definitions are often behind `$ref` / includes. Getting a quick overview of "which format is applied to which field, and what does it actually render as" requires jumping between both files for every case, which is slow and error-prone.

## Current behaviour

`clear-signing-tests-results.yml` already posts a `## Clear Signing Tests` comment on the PR. It contains a compact status table (one row per test case, one ✅/❌ column per implementation) and a `<details>` block with **Expected** / **Got** JSON only for cases that did not pass. A passing case is shown as a bare ✅, so the comment tells the reviewer that the descriptor and the test agree, but nothing about *what* they agree on.

## Proposal

Extend the results comment so that successful tests also surface their rendered output, joined with the descriptor's field definitions, in a form that can be read at a glance. For every affected descriptor, add a collapsed `<details>` section per test case with:

- the `intent` declared in the descriptor, the `interpolatedIntent` template if there is one, and the rendered `interpolatedIntent` produced by the runner
- one table row per rendered field: the field label, the `format` from the descriptor (plus the relevant `params`, e.g. `tokenPath`, `types`, `sources`), the `path`, and the rendered value

Mockup for one case:

<details>
<summary>✅ ekubo/calldata-Ve33Periphery · Stake tokens · <code>sourcify</code></summary>

**Intent:** `Stake` · **Template:** `Stake {amount}` · **Rendered:** `Stake 1,250 EKUBO`

| Label | Format | Params | Path | Rendered |
| --- | --- | --- | --- | --- |
| Amount | `tokenAmount` | `tokenPath: token` | `amount` | `1,250 EKUBO` |
| Beneficiary | `addressName` | `types: [eoa, wallet]`, `sources: [local, ens]` | `recipient` | `0x1F5e…05ee` |
| Lock duration | `duration` | | `duration` | `4 weeks` |
| Data | `raw` | | `data` | `0x` |

</details>

With that in the comment, a reviewer can check in a few seconds whether a raw `uint256` was left as `raw` instead of `tokenAmount`, whether an address field is missing `addressName`, or whether the interpolated intent reads naturally with real values, without opening either file.

## Implementation notes

- The rendered output is already available: `results.json` from `run-sourcify-tests` / `run-rust-tests` carries `cases[].rendered` with `intent`, `interpolatedIntent`, `owner` and `fields[{label, value}]`, and the `pr-context` artifact carries the test files. What is missing in the results job is the descriptor itself, so `detect-tests` should also copy the affected descriptors into `pr-context` (ideally resolved through `resolve-erc7730-includes.js`, so `$ref` and `includes` are already expanded).
- `renderedOutput.fields` is ordered like the descriptor's `fields`, per `erc7730-tests-v2.schema.json`, so rendered fields can be zipped with descriptor fields by position. Calldata-formatted fields nest recursively, so the table needs one level of indentation or a nested sub-table.
- The comment body is capped at 60000 characters. Each case should be collapsed by default, and above the cap the per-case sections should be dropped in favour of a link to the run, or written to `$GITHUB_STEP_SUMMARY` instead, keeping the status table intact.
- Only successful tests need this. Failing cases already show Expected / Got, so the new section should not duplicate them.

Related: #2957 (make CI more descriptive).

*Posted with Claude Code*

marcocastignoli:
Closes #2966.

## What changes

The Clear Signing Tests comment is now built by `.github/scripts/render-test-results.js` from the `pr-context` artifact, the runner artifacts and the test coverage report added in #2965, instead of inline in `clear-signing-tests-results.yml`. The github-script step only finds the pull request and upserts the comment. The script is checked out from the default branch, so a fork cannot change it.

For every case that passed, the comment adds a collapsed **Rendered fields** section: the rendered label and value of each field, nested calldata indented under its parent, followed by the display format of the function exactly as the descriptor declares it, with includes resolved and `$ref` left as they are. A reviewer checks a descriptor without opening the test file and the descriptor side by side. The status table, the Expected/Got blocks of failing cases, the missing-test block and the coverage block are unchanged.

## results.json

Two keys are added to the runner output, documented in `.github/test-runner-docs/README.md` and both examples:

- `descriptor`: the path of the descriptor under test, relative to the repository root.
- `cases[].format`: the `display.formats` key the implementation matched, the function signature or the EIP-712 primary type. Required on `pass` and `fail`, like `rendered`.

The renderer looks both up directly. It only follows descriptor paths of the shape `registry/<entity>/<name>.json`, so a runner cannot point outside the pull request context. Until the runners emit the keys, the section says that the runner did not report them; everything else in the comment keeps working from the artifact file name.

The pull request context now carries each affected descriptor with its includes resolved, stored under its repository path.

### DIFF
--- a/.github/scripts/render-test-results.js
+++ b/.github/scripts/render-test-results.js
@@ -23,8 +23,7 @@
  * test file and the descriptor side by side.
  *
  * Every input can come from a fork, so every string that lands in the comment
- * goes through text(), cell() or block(). To preview the comment locally, see
- * .github/test-runner-docs/README.md.
+ * goes through text(), cell() or block().
  */
 
 const fs = require('fs');
--- a/.github/scripts/render-test-results.js
+++ b/.github/scripts/render-test-results.js
@@ -1,0 +1,367 @@
+#!/usr/bin/env node
+/**
+ * Renders the "Clear Signing Tests" comment that clear-signing-tests-results.yml
+ * writes on a pull request, from the artifacts of one clear-signing-tests.yml
+ * run: the pr-context artifact (context.json, the test files and the resolved
+ * descriptors), the results__* artifacts of the runners and, when given, the
+ * test-coverage artifact (coverage.json, the report of check-test-coverage).
+ *
+ * Usage: node render-test-results.js --context <pr-context dir> --artifacts <results dir> [--coverage <test-coverage dir>]
+ *
+ * The run itself comes from the environment, because a fork cannot change the
+ * workflow_run event: RUN_URL, TESTED_SHA, COMMIT_URL, RUN_STARTED_AT and
+ * RUN_COMPLETED_AT. Writes the comment body to stdout. An empty body means
+ * that the workflow must remove the comment, because no descriptor was
+ * affected.
+ *
+ * The comment holds a status table with one column per implementation, the
+ * expected and the rendered output of every case that did not pass, and, for
+ * every case that passed, the rendered fields next to the display format that
+ * the descriptor declares for the function, as it is. The runner names the
+ * descriptor file in `descriptor` and the function in the `format` key of the
+ * case. The last part lets a reviewer check a descriptor without reading the
+ * test file and the descriptor side by side.
+ *
+ * Every input can come from a fork, so every string that lands in the comment
+ * goes through text(), cell() or block(). To preview the comment locally, see
+ * .github/test-runner-docs/README.md.
+ */
+
+const fs = require('fs');
+const path = require('path');
+const { parseArgs } = require('util');
+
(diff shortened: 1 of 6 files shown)


kuzdogan:
I'm closing this as it's superseded by #2997

Issue #2895, opened 2026-08-14
[Permit - common-dai-permit] `allowed` renders as a raw bool, and `false` silently revokes the allowance


kuzdogan:
### Related defects

The same defect class appears in these issues:

- #2819 (Tally Gitcoin governor, `Ballot.support` renders as `true`/`false` with no For or Against label)
- #2827 (Tally PoolTogether governor, the same defect)

The audit tracker #2852 groups these as category 6: a decision-critical value that the screen renders `raw`.

### Descriptor

`registry/permit/common-dai-permit.json`, added by the DAI-style permit PR 2855.

Two descriptors include this file:

- `registry/permit/eip712-permit-ethereum-dai.json`
- `registry/permit/eip712-permit-polygon-dai.json`

**Severity:** Medium

### Problem

The DAI `permit` function maps the boolean `allowed` to an allowance:

```solidity
uint wad = allowed ? uint(-1) : 0;
_approve(holder, spender, wad);
```

A value of `true` grants an allowance of `2^256 - 1`. A value of `false` sets the allowance to zero.

Reference: DAI on Ethereum, `0x6B175474E89094C44Da98b954EedeAC495271d0F`.

The descriptor renders `allowed` as a raw value:

```json
{ "path": "allowed", "label": "Unlimited spending", "format": "raw", "visible": "always" }
```

### Example

The screen shows one of these two lines:

```
Unlimited spending    true
Unlimited spending    false
```

Two problems follow.

1. The word `true` does not state the size of the allowance. The allowance is unlimited. That is the most dangerous property of a DAI permit, and the screen gives it as a bare boolean.
2. The line `Unlimited spending: false` does not state that the permit revokes the allowance. A signer can read that line as "no unlimited spending, so this message does nothing". The permit in fact sets the allowance to zero. The action and the screen do not agree.

### Remedy

**This defect is fixable today.** Use the `enum` format. The registry already uses this pattern in `registry/celo/calldata-celo_governance.json`:

```json
"metadata": { "enums": { "ProposalsVoteValue": { "0": "None", "1": "Abstain", "2": "No", "3": "Yes" } } }
```

```json
{ "path": "value", "label": "Vote", "format": "enum",
  "params": { "$ref": "$.metadata.enums.ProposalsVoteValue" }, "visible": "always" }
```

Apply the same shape to `allowed`:

```json
[...]

PR #2989, opened 2026-09-14
fix(lido): declare the address(0) owner sentinel, and state the scope of setApprovalForAll

### LINKED ISSUE #2720: Cyfrin Audit: [Lido - calldata-WithdrawalQueueERC721]
**Descriptor:** [registry/lido/calldata-WithdrawalQueueERC721.json](https://github.com/ethereum/clear-signing-erc7730-registry/blob/a2b33ffed92aab99cb9da2704610469a2588b520/registry/lido/calldata-WithdrawalQueueERC721.json)

**ERC-8176 descriptor hash:** `0xf96ad20c3880e2b196abf3da197cccaa475ca8eccea5e946b14bdef86dd79b09`

**1: `_owner = address(0)` means "msg.sender becomes the owner", but three request formats display it as a literal zero-address beneficiary**
**Severity:** Medium

**Description:** The WithdrawalQueue implementation (`0xE42C659dc09109566720EA8b2De186c2Be7D94D9`, behind the proxy at `0x889edC2eDab5f40e902b864aD4d7AdE8E412F9B1`) substitutes the caller: *"If `address(0)` is passed, `msg.sender` will be used as owner"*; `if (_owner == address(0)) _owner = msg.sender;`. The signer themselves receives the withdrawal NFT and the eventual ETH claim.

**Recommended mitigation:** `senderAddress` is the ERC-7730 `addressName` param for protocol-defined self-sentinels; per the spec, a field value equal to a listed address "is interpreted as the sender referenced by `@.from`", so the wallet renders **"Sender"** instead of the raw hex. The spec's own example uses exactly this configuration (`Sender` ← field value `0x0000...0000` with `senderAddress = ["0x0000...0000"]`), and the sentinel value matches this contract's documented `address(0)` to `msg.sender` substitution. A self-request then displays "Beneficiary: Sender", which is what executes on-chain.

```diff
// registry/lido/calldata-WithdrawalQueueERC721.json:77
+// In requestWithdrawals._owner, requestWithdrawalsWithPermit._owner,
+// and requestWithdrawalsWstETHWithPermit._owner; add the sentinel, matching line 77:
+"params": {
+"types": ["eoa", "wallet"],
+"sources": ["local", "ens"],
+"senderAddress": ["0x0000000000000000000000000000000000000000"]
+}
```

**2: `setApprovalForAll` display never conveys that approval covers ALL current and future unstETH withdrawal NFTs**
**Severity:** Medium

**Description:** Grants the operator authority to transfer **every** unstETH withdrawal NFT the signer holds now or mints later; each one a claim on withdrawn ETH.

**Recommended mitigation:** Change the affected descriptor entries as follows:

```diff
// registry/lido/calldata-WithdrawalQueueERC721.json:184
+"intent": "Allow operator to manage ALL your unstETH withdrawal NFTs"
+// optionally also relabel the boolean, e.g. "Grant access to all unstETH NFTs"
```

venugopalanvip:
Closes #2720 — both findings, 2 × Medium.

## Finding 1 — `_owner = address(0)` means the signer, not a burn address

`WithdrawalQueueERC721` substitutes the caller for a zero `_owner`:

```solidity
if (_owner == address(0)) _owner = msg.sender;
```

That line appears **twice** in the implementation — in `requestWithdrawals` and in
`requestWithdrawalsWstETH` — and the two permit variants `return` into them:

| function | where the substitution happens |
|---|---|
| `requestWithdrawals` | its own body |
| `requestWithdrawalsWstETH` | its own body |
| `requestWithdrawalsWithPermit` | `return requestWithdrawals(_amounts, _owner)` |
| `requestWithdrawalsWstETHWithPermit` | `return requestWithdrawalsWstETH(_amounts, _owner)` |

So all four share one `_owner` contract. The descriptor declared the
`senderAddress` sentinel on exactly **one** of them (`requestWithdrawalsWstETH`),
which is what makes this a plain internal inconsistency rather than a judgement
call — the file already contains its own correct answer. On the other three, a
signer sending a withdrawal to themselves read:

```
Beneficiary    0x0000000000000000000000000000000000000000
```

An address that can receive nothing, for a request whose proceeds are their own
ETH. The sentinel is now declared on all four.

**Verified against the deployment, not the audit text.** EIP-1967 slot on the
proxy `0x889edC2eDab5f40e902b864aD4d7AdE8E412F9B1` reads
`0xE42C659dc09109566720EA8b2De186c2Be7D94D9`, matching the implementation the
issue names; `name()` is `Lido: stETH Withdrawal NFT` and `symbol()` is
`unstETH`, so "unstETH NFT" in the intents is the protocol's own wording.

## Finding 2 — `setApprovalForAll` never said "all", and the bool never said which way

The call hands an operator authority over **every** unstETH NFT the signer holds
now or mints later, each one a claim on withdrawn ETH. The screen was:

```
Set unstETH operator approval
Operator    0x2aeD...1597
Approved    true
```

Neither line carries the scope, and a raw `true`/`false` leaves the signer to
infer the direction. `false` silently revokes — the same defect class as #2895
on the DAI permit `allowed` bool. The bool now renders through an enum, the shape
#2912 proposes there:

```
Manage ALL unstETH NFTs          ← Allow all unstETH NFTs / Revoke all unstETH NFTs
Operator    0x2aeD...1597
Access      Allow all / Revoke all
```

`intent` stays neutral at 23 chars because the static string cannot know the
bool; `interpolatedIntent` resolves the direction where the value is available.
I diverged from the audit's suggested `"Allow operator to manage ALL your unstETH
withdrawal NFTs"` only on length — it is 56 characters against the 30-character
intent limit.

## Fixtures — five new cases, and what they were worth finding out

Four pin the sentinel, **one per request function**. Each is that function's
existing fixture transaction with the `_owner` word zeroed and nothing else
touched — a 32-byte-for-32-byte substitution, so the RLP framing is untouched and
the encoding stays the one both runners already accept. The builder asserts that
no byte outside that word moved. Each sets `from`, which is what `@.from` resolves
against. The fifth covers the revoke branch of the new enum, which `Access` would
otherwise never exercise.

Three things worth recording from building them:

1. **`senderAddress` was almost untested.** Five descriptors declare it; before
   this PR exactly one entity had a fixture that reaches the sentinel value
   (`paraswap/calldata-AugustusSwapper-v6.2`, which is where I confirmed that the
   runners render it as `Sender` and that the test case must set `from`). Lido,
   `p2p/calldata-P2pSsvProxyFactory`, `uniswap/calldata-UniswapV3Router02` and
   `opencover/calldata-Quote` all declared it with nothing exercising it. This
   adds four cases.

2. **No real call reaches the sentinel.** I scanned the last 2,000 transactions
   to the queue: zero `requestWithdrawals` with `_owner = 0`. That is the reason
   the wrong rendering went unnoticed, not a reason to leave it — an integrator
   passing the documented convenience value gets the misleading screen.

3. **`interpolatedIntent` over an `enum` has descriptor precedent but no fixture.**
   `opencover/calldata-Quote` does it twice and has no `testsv2/` file, so the two
   `setApprovalForAll` cases here are the first to assert that rendering. If a
   runner substitutes the raw bool instead of the enum label, this PR is where
   that shows up — which is the point of asserting it.

## Checks

`check-selector-coverage.js` ✅ · `generate-index.js --validate` ✅ (buildable,
909/185, no new keys — the descriptor's deployments are unchanged) · both files
validate against the v2 descriptor and tests schemas · descriptor re-verified
byte-for-byte to contain no reformatting drift.

## One thing I did not change

On master, fixture `"Request Withdrawal - chain 1 (2)"` calls
`requestWithdrawalsWithPermit` (`0xacf41e4d`), not `requestWithdrawals` — its
description reads like a second case for the plain function. Both share the
intent `Request Withdrawal`, so nothing renders wrongly and I left it alone; it
did cost me one wrong fixture before I checked selectors rather than descriptions.


venugopalanvip:
Heads-up on a defect in this PR that the registry's CI cannot see.

The enum I added for `setApprovalForAll` uses boolean keys:

```json
"enums": { "operatorApproval": { "true": "Allow all", "false": "Revoke all" } }
```

That is valid per the v2 schema — "enumeration keys are the field values", any string allowed — and both reference runners pass on it, as the green checks here show. But `erc7730 calldata`, which generates the Ledger device descriptors, runs `int(ordinal)` on enum keys. `int("true")` raises, and the handler catching it discards the **whole descriptor**:

```
this branch : 0 device descriptors
master      : 11 device descriptors
```

So as it stands this PR would silently remove device support for the entire WithdrawalQueue contract, not just the format I touched. Neither runner exercises that path, which is why nothing here went red.

It is a library bug rather than a descriptor one — two descriptors already on master have the same problem and produce nothing today (`flyingtulip/calldata-PftNft.json`, `ekubo/calldata-MEVCaptureRouter.json`, the latter losing 168 device descriptors). Fix proposed in [LedgerHQ/python-erc7730#345](https://github.com/LedgerHQ/python-erc7730/pull/345); with it, this branch converts to 11 again.

I would rather not work around it here by renaming the keys to `"0"`/`"1"`, because the schema says keys are field values and `true`/`false` is what a bool field's values are — the two descriptors on master and #2912 all read it the same way. But if maintainers would rather this PR not wait on an upstream fix, say so and I will switch the keys and note why.

kuzdogan (approved):

PR #2997, opened 2026-09-17
ci: publish a visual test report for every descriptor pull request

kuzdogan:
## What a reviewer gets

The Clear Signing Tests comment gains one line: a link to a report page for the run, at `https://tests.erc7730.sourcify.dev/?pr=<n>&run=<id>`. The comment itself is unchanged.

The page shows:

- An overview of the affected descriptors: what changed, how many functions have a test, the pass and fail counts per implementation, and the Sourcify verification of the deployment.
- One card per test case. The expected screen from the test file sits on the left. Each implementation's rendered screen sits on the right, with the cells that differ marked. A case is labelled "pass everywhere", "implementations disagree" or "fails on all runners".
- The function parameters of each case, decoded from the raw transaction with the descriptor's format key as the ABI. Each argument, including the components of a struct, says whether the descriptor shows it, hides it with `visible: never` or the `excluded` list, or does not mention it at all. NatSpec comes from the verified source on Sourcify when there is one.
- The deployments of each descriptor, checked on Sourcify: verified, unverified, an address with a wrong EIP-55 checksum, a proxy with unverified implementations, or a verified ABI that lacks a function the descriptor formats. The last two are kept apart, because only the last one is a real mismatch.

## How it works

- The detect job of `clear-signing-tests.yml` copies each affected descriptor into `pr-context/descriptors/` with its includes resolved, the base version into `pr-context/base-descriptors/` when it exists, and writes a `changes` map into `context.json` that says whether the descriptor and its test file were added, modified, deleted or unchanged. The job keeps `contents: read`.
- `clear-signing-tests-results.yml` finds the pull request, builds `bundle.json` with `build-bundle.js` from the artifacts of the run, and runs `publish-report.sh`, which commits the bundle to the orphan branch `test-reports` as `pr/<n>/<run id>.json` and rewrites `pr/<n>/index.json` with `report-index.js`. The push retries up to three times when another run pushed first. The branch is created with a README on the first run. Every one of these steps continues on error, so the comment is always written.
- The viewer lives in `sourcifyeth/clear-signing-test-report` and reads the bundle from `raw.githubusercontent.com`. It also reads the test file at the head commit to decode the raw transactions, and asks Sourcify about chains, verification, ABIs and NatSpec.
- The bundle format is documented in `.github/test-runner-docs/bundle.md` as `schemaVersion` 1. It is the contract between this repository and the viewer.

## What changes for runners

Nothing is required. The runner guide now documents `rendered.fields` as an ordered array of `{label, value}` entries, which is what the test schema and both runners already use. The guide showed an object keyed by label. The guide also lists the optional case keys the bundle reads when a runner sends them: `warnings`, `format`, `chainId` and `durationMs`.

## Security

- The scripts run from a checkout of the default branch, so a fork cannot change them.
- The run id, head SHA, head repository and times come from the `workflow_run` event, not from an artifact.
- Only the results workflow has `contents: write`, and it pushes only to `test-reports`, a branch that holds no code any workflow runs. The token stays away from the dependencies: the scripts checkout persists no credentials, `npm ci` runs with `--ignore-scripts`, and no cache is restored.
- The bundle reads a matrix entry only when it names `registry/<entity>/<name>.json` with plain path segments, and only from inside the context directory.
- Every string in a bundle comes from a fork. The bundle copies it as data, and the viewer renders it as text and builds links only from values that match a strict pattern.
- The page can never pass or fail a pull request. The verdict stays in the runner jobs.

## What to check on the first run

Whether the organisation allows `contents: write` for the default token in this repository. If it does not, the publish step is the only failure, and the comment appears without the link. The `test-reports` branch already exists with two hand-published bundles for the runs of PR 2984 and PR 2994, so the viewer can be tried before the first automatic push.

## Related

Supersedes #2976 (rendered fields in the results comment). Two parts of that pull request live on here: the copy of the resolved descriptors into the pull request context, and the optional `format` key in the runner output, which the bundle reads when a runner sends it. The rendered-fields section of the comment is replaced by the link to the page.

Motivated by #2966 (rendered output in the PR comment).


.github/scripts/build-bundle.js
```diff
@@ -0,0 +1,410 @@
...
+  const descriptors = [];
+  for (const entry of matrix) {
+    const descriptorPath = entry.descriptor;
+    const entity = entry.entity;
+    const name = entry.descriptor_name;
+    const kind = name.startsWith('eip712-') ? 'eip712' : 'calldata';
+    const head = readJson(path.join(contextRoot, 'descriptors', descriptorPath));
+    const baseFile = path.join(contextRoot, 'base-descriptors', descriptorPath);
+    const base = fs.existsSync(baseFile) ? readJson(baseFile) : null;
+    const fixture = readJson(path.join(contextRoot, 'tests', `${entity}__${name}.tests.json`));
```
manuelwedler: **Path traversal from the fork-controlled context.** `entry.descriptor`, `entry.entity` and `entry.descriptor_name` come from `context.json`, which a fork can write freely (a fork can change the detect job in its own PR). They go into `path.join` unchecked, so `"descriptor": "../../../some/file.json"` makes this privileged job read any JSON file on the runner and publish it as `descriptors[].head` on a public branch.

I reproduced it locally:

```
context.json: {"matrix":[{"descriptor":"../../../secret/creds.json","entity":"x","descriptor_name":"calldata-x"}]}
bundle:       {"path":"../../../secret/creds.json","head":{"token":"SECRET-FROM-RUNNER"}}
```

A hosted runner holds little of value as JSON, so the impact is low today. But the fix is small and the header of this file promises that fork data is only copied as data. Suggestion: accept a matrix entry only if it matches a strict pattern, and check that the resolved path stays in the context directory.

```js
const SEGMENT = /^[A-Za-z0-9._-]+$/;
const inside = (root, file) => path.resolve(file).startsWith(path.resolve(root) + path.sep);
// skip the entry when entity/name do not match SEGMENT, when
// descriptorPath !== `registry/${entity}/${name}.json`, or when !inside(contextRoot, file)
```

The same check also protects against a `descriptor_name` that is not a string, which now throws in `name.startsWith` and drops the full report.

.github/workflows/clear-signing-tests-results.yml
```diff
@@ -63,15 +77,116 @@ jobs:
...
+        with:
+          node-version: "22"
+          cache: "npm"
+          cache-dependency-path: scripts/package-lock.json
+
+      - name: Install the dependencies of the scripts
+        if: steps.pr.outputs.found == 'true'
+        continue-on-error: true
+        working-directory: scripts
+        run: npm ci
```
manuelwedler: This job now has `contents: write` and it runs `npm ci` for the full dependency tree of the repository (the LangChain packages and all their transitive dependencies, only to get `viem`). Two small changes reduce what a compromised package can do here:

- `persist-credentials: false` on the checkout. The scripts checkout needs no git credentials later, and with the default the write token is on disk while the install scripts run.
- `npm ci --ignore-scripts`. `viem` needs no install scripts.

Optional: drop `cache: "npm"`. The install is small, and a cache restore in a job with a write token is one more input to trust.

Related to the comment on the `contents: write` line: I checked the branch protection. Only `master` is protected. Tags and all other branches are writable with this token, so it is worth keeping the token away from third-party code.

.github/scripts/publish-report.sh
```diff
@@ -0,0 +1,105 @@
...
+```
+
+The bundle format is documented in `.github/test-runner-docs/bundle.md` on `master`. The test report viewer reads the files from `raw.githubusercontent.com`.
+EOF
+}
+
+# Copies the bundle in and updates the index, then commits.
+apply() {
+  mkdir -p "$REPORT_DIR"
+  cp "$BUNDLE" "$REPORT_PATH"
```
manuelwedler: The size of the bundle is controlled by the fork: `head`, `base`, `dataProvider`, `expected`, `rendered` and `warnings` are copied from artifacts with no limit. So each push to a PR can commit a file of up to 100 MB to a branch of this repository, and a default `git clone` fetches that branch too.

Suggestion: refuse to publish a bundle above a limit (the two bundles on the branch are 16 KB and 88 KB, so 1 to 2 MB leaves a large margin). The comment then appears without the link, which is the failure mode that already exists.

```bash
MAX_BYTES="${MAX_BYTES:-2000000}"
if [ "$(wc -c < "$BUNDLE")" -gt "$MAX_BYTES" ]; then
  echo "bundle is larger than $MAX_BYTES bytes, not published" >&2; exit 1
fi
```

Not blocking, and possibly a follow-up: nothing prunes the branch. `bundle.md` says it can be recreated when it grows too large, but that is a manual step. A scheduled job that drops `pr/<n>/` of pull requests closed more than N days ago and squashes the branch would keep it bounded.

.github/workflows/clear-signing-tests.yml
```diff
@@ -294,6 +299,67 @@ jobs:
 
           cat pr-context/context.json
 
+      # The test report shows each descriptor as tested, the base version it
+      # replaces, and whether the pull request added or modified each file.
+      # The head checkout holds the head, base-scripts holds the base tree.
+      # Nothing here may fail the job: the runner jobs give the verdict, and
+      # the report shows a gap where a file could not be read.
+      - name: Add the descriptors and the changes to the pull request context
+        if: always()
```
manuelwedler: Could this be part of "Build the pull request context" instead of a second step? As it is, `context.json` has two writers: the first step writes it, and this step patches it with a read, a temp file and a `mv`. It also needs a second `TEST_MATRIX` default and the `-f pr-context/context.json` guard, which exist only because of the split.

In one step:

1. compute `CHANGES` first, with a fallback: `CHANGES=$(...) || CHANGES='{}'`
2. write `context.json` one time, with `base_sha` and `changes` in the same `jq -n` call
3. copy the test files
4. resolve the descriptors last. `resolve()` already turns a failure into a warning, so it cannot stop the step.

If you prefer to keep two steps, then this step needs `continue-on-error: true`. The comment says "Nothing here may fail the job", but the shell is `bash -e`, so a `jq` error in the `CHANGES=$(...)` assignment or in the `--argjson changes` call fails the detect job. The runner jobs that `need` it are then skipped, so a bug in this report-only step would stop the tests.

.github/scripts/build-bundle.js
```diff
@@ -0,0 +1,410 @@
...
+  const out = {};
+  if (!isObject(formats)) return out;
+  for (const key of Object.keys(formats)) {
+    const entry = { selector: null, primaryType: null, error: null, cases: [] };
+    if (kind === 'calldata') {
+      try {
+        entry.selector = toFunctionSelector(parseAbiItem(`function ${key.trim()}`));
+      } catch (error) {
+        entry.error = `not a function signature: ${reason(error)}`;
+      }
```
manuelwedler, line 156: Nit: `check-selector-coverage.js` already exports `selectorOf` and `testSelector`, with the same `parseAbiItem` / `parseTransaction` / `slice` logic and the same `reason` helper. If this file imports them, the coverage check and the report cannot drift apart on what "this test hits this format" means.

The same applies, less strongly, to `recommendationsOf`: `check-recommended-fields.js` tests `!('interpolatedIntent' in format)`, this file tests `== null || === ''`, so the comment and the report page can disagree for an empty string.

.github/test-runner-docs/README.md
```diff
@@ -28,26 +28,44 @@ Write a single `results.json` per descriptor to the working directory.
...
-        }
+        ]
       }
     }
   ]
 }
 ```
 
-`rendered` mirrors the shape of `expected` in the fixture — the same `{intent, owner, fields}` form applies to both calldata and EIP-712 cases.
+`rendered` mirrors the shape of `expected` in the fixture — the same `{intent, owner, fields}` form applies to both calldata and EIP-712 cases. `fields` is an ordered array of `{label, value}` entries, never an object keyed by label: labels can repeat when a field iterates over an array.
```
manuelwedler, line 68: Nit: `build-bundle.js` reads four optional keys from each case (`warnings`, `format`, `chainId`, `durationMs`), but the Fields table below does not list them, so a runner author cannot know they exist. A row for each, marked optional, would close the gap.

Also, the PR description says the bundle reads an optional `descriptor` key from the runner output. I found no code that reads it. Either the description or the script is out of date.

.github/scripts/build-bundle.js
kuzdogan: Fixed in d148892. `isMatrixEntry` accepts an entry only when `entity` and `descriptor_name` match `^(?!\.+$)[A-Za-z0-9._-]+$` and `descriptor` equals `registry/<entity>/<name>.json`. The three files it then reads are also checked to resolve inside the context directory. Anything else is skipped with a warning, so a bad entry no longer drops the whole report. Tested locally with your `../../` entry: skipped, the real entry next to it still builds.

.github/workflows/clear-signing-tests-results.yml
kuzdogan: Done, all three: `persist-credentials: false` on the scripts checkout, `npm ci --ignore-scripts`, and the npm cache is gone. Thanks for checking the branch protection, I noted it in the PR body.

.github/workflows/clear-signing-tests.yml
kuzdogan: You are right about `bash -e`. I took the one-step option. "Build the pull request context" now computes `CHANGES` first with `|| CHANGES='{}'`, writes `context.json` once with `base_sha` and `changes` in the same `jq -n` call, copies the test files, and resolves the descriptors last. The second step, the second `TEST_MATRIX` default, the temp file and the `-f` guard are gone.

The workflow does not run on this PR, because only `.github` files change. So I ran the step locally with `bash -eo pipefail`, a real Flare matrix entry and a hostile one: exit 0, head and base resolved, `changes` correct.

.github/scripts/build-bundle.js
kuzdogan: Done. `build-bundle.js` now imports `selectorOf`, `testSelector` and `reason` from `check-selector-coverage.js`. `inputOf` still parses the transaction for the chain, the recipient and the value, but the selector comes from `testSelector`. `recommendationsOf` uses `!('interpolatedIntent' in format)` like the check.

.github/test-runner-docs/README.md
kuzdogan: Added four rows to the Fields table, all optional: `warnings`, `format`, `chainId`, `durationMs`, with a pointer to `bundle.md`. The `descriptor` key was the description being out of date, not the script. Only `format` landed. The PR body now says so.

manuelwedler (approved):
lgtm

